PR #1, opened 2017-10-01
Refine searching and filtering behavior (UIREQ-11)

Baroquem:


Baroquem:
It works fine on my system, so I'd suspect a server-side issue. I'm running the latest testing-backend VM; earlier versions may not support the CQL queries.

zburke:
Search issues identified above were all related to https://issues.folio.org/browse/STCOM-84.

PR #2, opened 2017-10-09
Merge bug fixes for UIREQ-20

Baroquem:


mkuklis:
@Baroquem this looks good! The `console.log` should be probably removed and then it should be ready to go.

Baroquem:
True, thanks for pointing that out.

MikeTaylor:
Note, though, that the `console.log`s are not part of this PR.

So they should probably go; but their continuing existence should not be a reason to delay merging this PR. (As, indeed, it hasn't been delayed.)

PR #5, opened 2017-10-19
Add an edit view for requests (completes UIREQ-8)

Baroquem:


zburke:
`yarn lint` does not run cleanly, in some case resulting in console errors.

PR #43, opened 2018-02-12
Fix tests

mkuklis:


zburke:
Thanks, @mkuklis, for fixing these. I should have run the test-suite after I made the changes and then I would have caught this; my bad.

mkuklis:
@zburke no worries Zak. I was going through different tests to get the CI green again. They are passing for me locally but it looks like they are still failing on CI :(

PR #92, opened 2018-06-29
UIREQ-58 and UIREQ-67: Cancel Requests 

doytch:
Adds the ability to cancel requests via a `CancelRequestDialog` component. It can be pulled up using a new actions dropdown on the `ViewRequest` pane header.


doytch:
@Baroquem added ya since you've done work in here and it _looked_ like you added the whole process behind building out `fullRequestDetail`. Editing (or cancelling) a request wasn't showing the updated request when the detail view was open, so I added a `getDerivedStateFromProps` to carry over the updated request to `fullRequestDetail` to prompt a re-render when necessary.

PR #121, opened 2018-09-12
STCOM-337

rasmuswoelk:
Updated MCL column label's sentence structure to be consistent with the overall design of FOLIO.

https://issues.folio.org/browse/STCOM-337


rasmuswoelk:
I realized that I made some mistakes here. I'm going to fix them now and push some changes.

rasmuswoelk:
Made some changes. Added translated strings and made sure that sorting still works as intended.

PR #124, opened 2018-09-14
 Lint JS

cherewaty:
- Moves to `eslint-config-stripes` `3.2`
- Cleans up `peerDependencies` warnings

PR #127, opened 2018-09-17
Fix malformed JSX

cherewaty:
Bug fix for https://github.com/folio-org/ui-requests/pull/124

### DIFF
--- a/index.js
+++ b/index.js
@@ -23,7 +23,10 @@ class RequestsRouting extends React.Component {
       <div>
         <h2>Uh-oh!</h2>
         <p>
-          {`How did you get to <tt>${this.props.location.pathname}</tt>?`}
+          How did you get to
+          {' '}
+          <tt>{this.props.location.pathname}</tt>
+          ?
         </p>
       </div>
     );


MikeTaylor:
Why was this malformed?

zburke:
ESLint gets cranky unless JSX elements are all on their own lines. "Malformed" may be a little strong, though maybe from ESLint's point of view it's the right word.

MikeTaylor:
Ugh. It always feels like a violation to me when we make code longer, uglier and hard to understand in order to satisfy a "style" tool. Tail wags dog.

zburke:
To be fair, I believe there is potentially an XSS issue with React having HTML tags embedded in interpolated strings, which this also avoids. Click the "show outdated" link in [this similar PR](https://github.com/folio-org/stripes-core/pull/415#pullrequestreview-155962685) for more details. 

If I squint, I can just barely see the wisdom of ESLint's blanket policy of no markup in interpolated strings because it's just too dangerous if you do it wrong. But in this case, with a hard-coded `<tt>` and no attributes, we're obviously _not_ doing it wrong so, yeah, changing it rubs.

PR #136, opened 2018-10-02
Use stripes 1.0

cherewaty:
Uses new `@folio/stripes` package.


src/ViewRequest.js
```diff
@@ -209,6 +212,11 @@ class ViewRequest extends React.Component {
     });
   }
 
+  craftLayerUrl(mode) {
```
ttamj: Duplicate of https://github.com/folio-org/ui-requests/pull/135?

cherewaty: Merged.

PR #169, opened 2018-11-19
UIREQ-147: Apply internationalization to all hardcoded strings

MaksymDryha:
According to [UIREQ-147,](https://issues.folio.org/browse/UIREQ-147) all the hardcoded strings should be internationalized using the react-intl pattern constructs outlined [here](https://github.com/folio-org/stripes/blob/master/doc/i18n.md).


src/ViewRequest.js
```diff
@@ -291,8 +291,8 @@ class ViewRequest extends React.Component {
         lastMenu={detailMenu}
         actionMenuItems={!isRequestClosed ? [{
           id: 'clickable-edit-request',
-          title: 'Edit Request',
-          label: 'Edit',
+          title: formatMessage({ id: 'ui-requests.actions.editRequest' }),
```
mkuklis: @MaksymDryha is there a reason we can't use `<FormattedMessage>` here?

src/index.js
```diff
@@ -22,9 +23,9 @@ class RequestsRouting extends React.Component {
...
     return (
       <div>
-        <h2>Uh-oh!</h2>
+        <h2><FormattedMessage id="ui-requests.routingErrorOops" /></h2>
         <p>
-          How did you get to
+          <FormattedMessage id="ui-requests.routingError" />
           {' '}
           <tt>{this.props.location.pathname}</tt>
           ?
```
anastasiaHahanova: Please, add this to routingError as value
` {' '}<tt>{this.props.location.pathname}</tt> ? `

src/ViewRequest.js
MaksymDryha: The reason I didn't use `<FormattedMessage>` here was that `actionMenuItems` is an array with more than one title string. Wrapping `<Pane>` with `<FormattedMessage>`, would only allow us to translate one string so I decide to use `formatMessage` separetly for each title.
But based on @cherewaty's comment about `title`, this is no longer relevant. I removed those attributes.

PR #197, opened 2018-12-20
UIREQ-129: BigTest initial setup

Godlevskyi:
## Purpose
https://issues.folio.org/browse/UIREQ-129

We want to add both unit test coverage to the requests application, and also add the ability to start the application in mirage mode so that development can be done without having to actually be connected to any kind of server environment.

## Approach

This adds BigTest and Mirage to the requests app using the emerging standard bigTest setup of models/serializers/factories going into the tests/bigTest directory.

Endpoints were being setup against an existing API.

In addition to the endpoints, the default scenario was populated with about 20 requests instance records 

## Strategy

There is a lot of work to be done in creating the initial stub that lets you achieve the very basic functionality. After that, we can proceed incrementally to implement features of the harness in order to facilitate more comprehensive tests.

As such, the strategy was to stand up just enough of the endpoints to be able to both view the application and write a basic test that loaded a list of users over the network, made sure that all the instances rendered properly, and that clicking on a item in the list loaded the corresponding request record.

In this PR only the _read only_ endpoints for most entities have been implemented.


test/bigtest/network/factories/addresse.js
```diff
@@ -0,0 +1,18 @@
+import { Factory, faker } from '@bigtest/mirage';
```
zburke: This file should be named `address.js`. Likewise for `.../network/models/addresse.js`. Everything else looks great!

Godlevskyi: Done!

PR #204, opened 2018-12-31
Add feature from fees/fines 

DGB-UNAM-FOLIO:
We are adding:

 
 * Add feature enforce manual patron blocks. Refs [UIU-675](https://issues.folio.org/browse/UIU-675).


DGB-UNAM-FOLIO:
Please, @mkuklis, @zburke and @skoczko, help me verify and merge the code of this function for fees/fines.
Thank you.

src/RequestForm.js
```diff
@@ -803,7 +845,7 @@ export default stripesForm({
   form: 'requestForm',
   asyncValidate,
   asyncBlurFields: ['item.barcode', 'requester.barcode'],
-  navigationCheck: true,
+  navigationCheck: false,
```
zburke: This change allows a user to navigate away from a dirty form without being prompted to save their changes. That seems incorrect. Can you explain why it changed?

DGB-UNAM-FOLIO: This is because in the modal "Patron Blocked from requesting", when pressing the "Block details" button, the validation modal (NavigationCheck) appears, which does not allow to see the list and user details immediately.

In addition to not showing the list of users of the issue UIU-675.

PR #215, opened 2019-01-23
Fix fees/fines 

DGB-UNAM-FOLIO:
* Fix expiration date for patron blocks. Fix [UIU-792](https://issues.folio.org/browse/UIU-792).


src/RequestForm.js
```diff
@@ -270,7 +279,8 @@ class RequestForm extends React.Component {
   }
 
   findUser(barcode) {
-    const blocks = get(this.props.parentResources, ['patronBlocks', 'records'], []).filter(b => b.requests === true);
+    let blocks = get(this.props.parentResources, ['patronBlocks', 'records'], []).filter(b => b.requests === true);
```
mkuklis: @DGB-UNAM-FOLIO it looks like we repeat the same code (getting and filtering blocks) in `findUser`, `onItemClick`, `render` and couple other places. Would it be possible for you to extract it into a separate method?

PR #234, opened 2019-02-01
Turn on navigation check on request form

mkuklis:
I wish all form/navigation related issues were that simple...

### DIFF
--- a/src/RequestForm.js
+++ b/src/RequestForm.js
@@ -858,7 +858,7 @@ export default stripesForm({
   form: 'requestForm',
   asyncValidate,
   asyncBlurFields: ['item.barcode', 'requester.barcode'],
-  navigationCheck: false,
+  navigationCheck: true,
   enableReinitialize: true,
   keepDirtyOnReinitialize: true,
 })(injectIntl(RequestForm));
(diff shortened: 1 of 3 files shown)


src/RequestForm.js
```diff
@@ -858,7 +858,7 @@ export default stripesForm({
   form: 'requestForm',
   asyncValidate,
   asyncBlurFields: ['item.barcode', 'requester.barcode'],
-  navigationCheck: false,
+  navigationCheck: true,
```
zburke: This one was bugging me and it took me a while to remember why. See the conversation about it under https://github.com/folio-org/ui-requests/pull/204 / https://issues.folio.org/browse/UIU-675. 

Maybe we should leave it alone for now?

mkuklis: Oh no. Your intuition was right @zburke :(. It looks like what was done in #204 is a work around but not a real fix. I will take a quick look at it today. If it's really impossible to fix it we should leave it alone for now.

mkuklis: @zburke I made a small change which will reinitialize the form when the `Block details` button is clicked. That should allow us to leave the request form without seeing the navigation check modal. This is not ideal but I think it's still better than completely turning the check off.

zburke: Looks good. I trust your judgement here.

PR #272, opened 2019-02-18
Fix saving empty hold shelf expiration date. Fixes UIREQ-206

mkuklis:
https://issues.folio.org/browse/UIREQ-206


mkuklis:
Thank you for reviewing it @Baroquem. The strings in the constants are used to perform the query against the backend and are not used for the data display. Perhaps I should rename that variable to something else to make it clearer? What do you think?

Baroquem:
No, I think the variable name is ok ... I just got mixed up because it behaves differently from the other constants. Maybe you could just add a comment to that effect.

PR #273, opened 2019-02-19
Update circulation v7.0 and request-storage v3.0 OKAPI interfaces. Part of UIREQ-214

mkuklis:
https://issues.folio.org/browse/UIREQ-214

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,9 +27,9 @@
     "queryResource": "query",
     "okapiInterfaces": {
       "cancellation-reason-storage": "1.1",
-      "circulation": "3.4 4.0 5.0 6.0 7.0",
+      "circulation": "7.0",
       "inventory": "6.0 7.0 8.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -27,9 +27,9 @@
     "queryResource": "query",
     "okapiInterfaces": {
       "cancellation-reason-storage": "1.1",
-      "circulation": "3.4 4.0 5.0 6.0",
+      "circulation": "3.4 4.0 5.0 6.0 7.0",
```
marcjohnson-kint: This change appears to include the changes for sending date and time for the two expiration date properties.

That format is not compatible with the older versions. How can this offer dual support, given that it is reliant upon the breaking changes for requests?

mkuklis: @marc-johnson that's a good catch. I think we can only include `"circulation": "7.0"`

PR #293, opened 2019-03-01
Fix item status. Fixes UIREQ-222

mkuklis:
https://issues.folio.org/browse/UIREQ-222

### DIFF
--- a/src/ItemDetail.js
+++ b/src/ItemDetail.js
@@ -11,7 +11,7 @@ const ItemDetail = ({ item, loan, requestCount }) => {
   }
 
   const recordLink = item.barcode ? <Link to={`/inventory/view/${item.instanceId}/${item.holdingsRecordId}/${item.id}`}>{item.barcode}</Link> : '';
-  const status = get(item, ['status', 'name'], '');
+  const status = get(item, 'status.name') || get(item, 'status');
   const contributor = get(item, ['contributorNames', '0', 'name'], '-');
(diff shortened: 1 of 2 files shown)


src/ItemDetail.js
```diff
@@ -11,7 +11,7 @@ const ItemDetail = ({ item, loan, requestCount }) => {
   }
 
   const recordLink = item.barcode ? <Link to={`/inventory/view/${item.instanceId}/${item.holdingsRecordId}/${item.id}`}>{item.barcode}</Link> : '';
-  const status = get(item, ['status', 'name'], '');
+  const status = get(item, 'status.name') || get(item, 'status');
```
zburke: Wait wait wait wait wait wait wait. How is it possible that sometime's a thing's status is sometimes a string and sometimes an object? @marcjohnson-kint, do you have a sense of what's happening on the back-end that would cause this kind of contortion on the front-end? 

Aside, do we we have a table enumerating the possible values somewhere? This feels like the kind of value we'd want to translate before we display it.

marcjohnson-kint: @zburke This was discussed during refinement on Friday.

It was believe that the the item status was coming from different places depending upon when in the process. Before a request is created, the item status comes from the item fetched directly. Afterwards, and when viewing etc, it comes from the request's representation of an item.

And this workaround was intended until the breaking change could be made to make them consistent. And this was going to wait till after the circulation interface is split.

However, it appears that belief might be incorrect, as the request's representation of the item does not contain a status in the [schema](https://github.com/folio-org/mod-circulation/blob/ce601d5c62dcdbba1d7847a0b356e16836012158/ramls/request.json#L69), so I am now a little confused.

zburke: Works for me.

PR #311, opened 2019-03-13
Add field level validtions to pickupservice point and delivery addres…

aditya-matukumalli:
…s, refs UIREQ-227

The additional bugs mentioned in the story should be resolved as well, as the validations shouldn't allows us in the first place to submit a request without a service point or a delivery address.

### DIFF
--- a/src/UserForm.js
+++ b/src/UserForm.js
@@ -42,7 +42,7 @@ class UserForm extends React.Component {
     this.connectedProxyManager = props.stripes.connect(ProxyManager);
   }
 
-  requireServicepoint = value => (value ? undefined : <FormattedMessage id="ui-requests.errors.selectItem" />);
+  requireServicePoint = value => (value ? undefined : <FormattedMessage id="ui-requests.errors.selectItem" />);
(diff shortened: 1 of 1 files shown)


src/UserForm.js
```diff
@@ -55,6 +58,7 @@ class UserForm extends React.Component {
         component={Select}
         fullWidth
         onChange={onChangeAddress}
+        validate={this.requireDeliveryAddress}
```
zburke: if a selection is required, shouldn't we pass `required` to its `<Field>` as well?

src/UserForm.js
```diff
@@ -75,6 +79,7 @@ class UserForm extends React.Component {
         label={<FormattedMessage id="ui-requests.requester.pickupServicePoint" />}
         component={Select}
         fullWidth
+        validate={this.requireServicepoint}
```
zburke: if a selection is required, shouldn't we pass `required` to its `<Field>` as well?

also, minor, `requireServicePoint` is a more consistent name for this function.

src/UserForm.js
aditya-matukumalli: Good catch. Will add it.

PR #320, opened 2019-04-02
Fix double detail pane view when clicking back. Part of UIREQ-231

mkuklis:
This PR refactors ViewRequest render to avoid double pane rendering. 

https://issues.folio.org/browse/UIREQ-231

### DIFF
--- a/src/ViewRequest.js
+++ b/src/ViewRequest.js
@@ -258,8 +258,6 @@ class ViewRequest extends React.Component {
     // TODO: Internationalize this
     const isRequestClosed = requestStatus.startsWith('Closed');
 
-    console.log('isRequestClosed', isRequestClosed);
-
     return (
       <PaneMenu>
         {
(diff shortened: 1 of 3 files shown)


src/ViewRequest.js
```diff
@@ -269,7 +286,18 @@ class ViewRequest extends React.Component {
...
+  renderRequest(request) {
+    const {
+      stripes,
+    } = this.props;
+
+    const patronGroupName = this.getPatronGroupName(request);
+    const getPickupServicePointName = this.getPickupServicePointName(request);
+    const requestStatus = get(request, ['status'], '-');
+    // TODO: Internationalize this
+    const isRequestClosed = requestStatus.startsWith('Closed');
```
zburke: There's no i18n work here; this is reading a value presented from the server.

mkuklis: That's good to know!

PR #335, opened 2019-04-23
Add ability to create a request without an item barcode. Part of UIREQ-253.

mkuklis:
`On Order` items don't have barcodes but we still need to be able to create requests for them when coming from the inventory module. 

https://issues.folio.org/browse/UIREQ-253

### DIFF
--- a/src/ItemDetail.js
+++ b/src/ItemDetail.js
@@ -14,7 +14,7 @@ const ItemDetail = ({ item, loan, itemId, requestCount }) => {
   const status = get(item, 'status.name') || get(item, 'status');
   const contributor = get(item, ['contributorNames', '0', 'name'], '-');
   const positionLink = item.barcode ? <Link to={`/requests?filters=requestStatus.Open%20-%20Awaiting%20pickup%2CrequestStatus.Open%20-%20In%20transit%2CrequestStatus.Open%20-%20Not%20yet%20filled&query=${item.barcode}&sort=Request%20Date`}>{requestCount}</Link> : '-';
-  const itemLabel = item.barcode ? 'ui-requests.item.id' : 'ui-requests.item.barcode';
+  const itemLabel = item.barcode ? 'ui-requests.item.barcode' : 'ui-requests.item.id';
 
   return (
     <React.Fragment>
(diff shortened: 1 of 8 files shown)


src/ItemDetail.js
```diff
@@ -6,20 +6,21 @@ import { FormattedDate, FormattedMessage } from 'react-intl';
...
     return <FormattedMessage id="ui-requests.actions.loading" />;
   }
 
-  const recordLink = item.barcode ? <Link to={`/inventory/view/${item.instanceId}/${item.holdingsRecordId}/${itemId}`}>{item.barcode}</Link> : '';
+  const recordLink = item ? <Link to={`/inventory/view/${item.instanceId}/${item.holdingsRecordId}/${itemId}`}>{item.barcode || item.id}</Link> : '-';
   const status = get(item, 'status.name') || get(item, 'status');
   const contributor = get(item, ['contributorNames', '0', 'name'], '-');
-  const positionLink = item ? <Link to={`/requests?filters=requestStatus.Open%20-%20Awaiting%20pickup%2CrequestStatus.Open%20-%20In%20transit%2CrequestStatus.Open%20-%20Not%20yet%20filled&query=${item.barcode}&sort=Request%20Date`}>{requestCount}</Link> : '-';
+  const positionLink = item.barcode ? <Link to={`/requests?filters=requestStatus.Open%20-%20Awaiting%20pickup%2CrequestStatus.Open%20-%20In%20transit%2CrequestStatus.Open%20-%20Not%20yet%20filled&query=${item.barcode}&sort=Request%20Date`}>{requestCount}</Link> : '-';
+  const itemLabel = item.barcode ? 'ui-requests.item.id' : 'ui-requests.item.barcode';
```
Baroquem: I think you have the labels mixed up here.

src/RequestForm.js
```diff
@@ -689,7 +697,7 @@ class RequestForm extends React.Component {
                                     ref={this.itemBarcodeRef}
                                     onInput={this.onItemClickDebounce}
                                     onKeyDown={e => this.onKeyDown(e, 'item')}
-                                    validate={this.requireItem}
+
```
Baroquem: If validation isn't needed on this field anymore, then the requireItem function can be removed altogether (and maybe the translation string that it uses).

src/ItemDetail.js
mkuklis: @Baroquem good catch! Thank you!

src/RequestForm.js
mkuklis: Good idea @Baroquem! I will keep the translation around because it's used in couple different places.

PR #389, opened 2019-08-12
Prevent move above page. Part of UIREQ-303

mkuklis:
https://issues.folio.org/browse/UIREQ-303


mkuklis:
@zburke it really is :(. There is some coverage in BigTest (it should cover moving request and also this scenario) but I agree with you it would be great to have couple integration tests too.

I would like to get this to Cate now so she can do some more testing and make sure this actually works as expected but perhaps after that we could add couple integration tests?

PR #399, opened 2019-09-05
Merge changes for UIREQ-320

Baroquem:


zburke:
The failing unit tests on this PR appear to be the result of the `page` requests that are randomly created by the request factory, causing new, non-page requests to act differently when placed on items with existing page requests (i.e. when the "can't move a request above a page" dialog is shown). Removing `page` as an option in the factor, my tests pass consistently. 

Is this the right fix, or should this test just be more careful about the data it mocks so that pages are not present in its test scenarios?

PR #434, opened 2019-11-13
Display cancellation reason on cancelled request. Part of UIREQ-364

mkuklis:
https://issues.folio.org/browse/UIREQ-364

### DIFF
--- a/src/routes/RequestsRoute.js
+++ b/src/routes/RequestsRoute.js
@@ -168,6 +168,11 @@ class RequestsRoute extends React.Component {
       path: 'circulation/requests-report/expired-holds',
       fetch: false,
     },
+    cancellationReasons: {
+      type: 'okapi',
+      path: 'cancellation-reason-storage/cancellation-reasons',
+      records: 'cancellationReasons',
+    },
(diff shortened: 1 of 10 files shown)


src/routes/RequestsRoute.js
```diff
@@ -563,6 +568,7 @@ class RequestsRoute extends React.Component {
     const patronGroups = (resources.patronGroups || {}).records || [];
     const addressTypes = (resources.addressTypes || {}).records || [];
     const servicePoints = (resources.servicePoints || {}).records || [];
+    const cancellationReasons = get(resources, 'cancellationReasons.records', []);
```
zburke: Can we keep the same structure for all resources here, i.e. use 
```
get(resources, 'resourceName.records', []);
```
or 
```
(resources.resourceName || {}).records || [];
```
across the board?

mkuklis: Yes I was thinking the same but felt it's out of this PR. I think it would be good to do all sort of cleanup in ui-requests and ui-inventory. I will update these now.

PR #444, opened 2019-11-21
UIREQ-369: disallow editing fulfillment preference on requests with Awaiting delivery/pickup status

MaksymDryha:
### Purpose
Disallow editing of fulfillment preference when a request is in Open - Awaiting delivery/pickup status

[Related story](https://issues.folio.org/browse/UIREQ-369)

### DIFF
--- a/src/UserForm.js
+++ b/src/UserForm.js
@@ -8,6 +8,12 @@ import { Field } from 'redux-form';
 import { Col, KeyValue, Row, Select } from '@folio/stripes/components';
 import { ProxyManager } from '@folio/stripes/smart-components';
 import { getFullName, userHighlightBox } from './utils';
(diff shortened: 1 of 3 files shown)


src/UserForm.js
```diff
@@ -104,11 +110,11 @@ class UserForm extends React.Component {
       request,
     } = this.props;
 
-
     const id = user.id;
     const name = getFullName(user);
     const barcode = user.barcode;
-    const editMode = get(request, 'id');
+    const editMode = !!request;
```
zburke: Maybe rename `editMode` to `isEditable`? "mode" implies something other than a boolean to me.

MaksymDryha: @zburke, makes sense, done!

PR #450, opened 2019-11-26
UIREQ-382 - FE: Print pick slips for service point on demand

maximdidenkoepam:
# Purpose
Get all pick slips in the queue that are associated with current service point to print

# Link
https://issues.folio.org/browse/UIREQ-382

# Screenshots
Screen Shot 2019-11-27 at 16 25 23
Screen Shot 2019-11-27 at 16 25 33

### DIFF
--- a/src/components/PrintButton/PrintButton.js
+++ b/src/components/PrintButton/PrintButton.js
@@ -64,7 +64,10 @@ class PrintButton extends React.Component {
           <div ref={this.printContentRef}>
             {dataSource.map(source => {
               return (
-                <div style={{ 'page-break-before': 'always' }}>
+                <div
+                  key={source['request.requestID']}
+                  style={{ pageBreakBefore: 'always' }}
+                >
                   <ComponentToPrint
                     template={template}
                     dataSource={source}
--- a/src/components/PrintButton/PrintButton.js
+++ b/src/components/PrintButton/PrintButton.js
@@ -62,19 +62,17 @@ class PrintButton extends React.Component {
         />
         <div className={css.hiddenContent}>
           <div ref={this.printContentRef}>
-            {dataSource.map(source => {
-              return (
-                <div
-                  key={source['request.requestID']}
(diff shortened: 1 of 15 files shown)


src/components/PrintButton/PrintButton.js
```diff
@@ -0,0 +1,85 @@
...
+          content={this.getContent}
+          removeAfterPrint
+          trigger={this.renderTriggerButton}
+          onAfterPrint={onAfterPrint}
+          onBeforePrint={onBeforePrint}
+        />
+        <div className={css.hiddenContent}>
+          <div ref={this.printContentRef}>
+            {dataSource.map(source => {
+              return (
```
mkuklis: @maximdidenkoepam I wonder if we need the `return` here. Could we just get rid of `{`.

src/routes/RequestsRoute.js
```diff
@@ -173,6 +177,12 @@ class RequestsRoute extends React.Component {
       path: 'cancellation-reason-storage/cancellation-reasons',
       records: 'cancellationReasons',
     },
+    staffSlips: {
```
marcjohnson-kint: Is this a new backend interface dependency, it appears to be a new endpoint for this module, and there doesn’t appear to be an entry in the package.json for it?

src/routes/RequestsRoute.js
```diff
@@ -173,6 +177,12 @@ class RequestsRoute extends React.Component {
...
+      type: 'okapi',
+      records: 'staffSlips',
+      path: 'staff-slips-storage/staff-slips?limit=100',
```
marcjohnson-kint: Also, I was of the understanding that new slip templates weren’t going to use the older staff slips templates and instead use the template engine, did I misunderstand that?

src/routes/RequestsRoute.js
zburke: I agree, @marcjohnson-kint ; it looks like we need to add a dependency on mod-circulation-storage.

PR #452, opened 2019-12-01
UIREQ-382 - Print pick slips for service point on demand

maximdidenkoepam:
# Purpose
Get all pick slips in the queue that are associated with current service point to print

# Link
https://issues.folio.org/browse/UIREQ-382

# Screenshots
Screen Shot 2019-11-27 at 16 25 23
Screenshot 2019-12-01 at 22 23 59

### DIFF
--- a/src/routes/RequestsRoute.js
+++ b/src/routes/RequestsRoute.js
@@ -220,10 +220,10 @@ class RequestsRoute extends React.Component {
         GET: PropTypes.func,
       }),
       pickSlips: PropTypes.shape({
-        GET: PropTypes.func,
+        GET: PropTypes.func.isRequired,
       }).isRequired,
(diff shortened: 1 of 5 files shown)


src/routes/RequestsRoute.js
```diff
@@ -212,6 +219,12 @@ class RequestsRoute extends React.Component {
       staffSlips: PropTypes.shape({
         GET: PropTypes.func,
       }),
+      pickSlips: PropTypes.shape({
+        GET: PropTypes.func,
+      }),
```
VictorSoroka1: I think we can be sure that this one is required as well as the presence of the GET method, right? If so please add these for _pickSlips_ and _currentServicePoint_.

PR #465, opened 2019-12-10
UIREQ-385: Fix bug with wrong menus when duplicating request

Vladyslav-Velytskyi:
# Purpose
Show the Fulfillment preference menu with Hold shelf selected and the menu for selecting the service point for pickup when user logs in from user(not diku_admin) and trying to duplicate the request

# Link
https://issues.folio.org/browse/UIREQ-385

# Screenshots
## Before
screenshot-1

## After
bRkWMJdHQm

### DIFF
--- a/src/RequestForm.js
+++ b/src/RequestForm.js
@@ -18,6 +18,8 @@ import {
   cloneDeep,
   defer,
   unset,
+  pick,
+  isBoolean,
 } from 'lodash';
 
 import { Pluggable } from '@folio/stripes/core';
@@ -114,9 +116,8 @@ class RequestForm extends React.Component {
   constructor(props) {
     super(props);
 
-    const { request, initialValues } = props;
+    const { request } = props;
     const { requester, item, loan } = (request || {});
-    const { deliveryAddressTypeId } = initialValues;
 
     this.state = {
       accordions: {
@@ -125,16 +126,11 @@ class RequestForm extends React.Component {
         'requester-info': true,
       },
       proxy: {},
-      hasDelivery: false,
-      requestPreferencesLoaded: false,
-      defaultDeliveryAddressTypeId: '',
-      defaultServicePointId: '',
-      selectedDelivery: isDelivery(initialValues),
-      selectedAddressTypeId: deliveryAddressTypeId,
       selectedItem: item,
       selectedUser: requester,
       selectedLoan: loan,
       blocked: false,
+      ...this.getDefaultRequestPreferences(),
     };
 
     this.connectedCancelRequestDialog = props.stripes.connect(CancelRequestDialog);
@@ -188,7 +184,7 @@ class RequestForm extends React.Component {
       // eslint-disable-next-line react/no-did-update-set-state
       this.setState({
         selectedAddressTypeId: initialValues.deliveryAddressTypeId,
-        selectedDelivery: isDelivery(initialValues),
+        deliverySelected: isDelivery(initialValues),
         selectedItem: request.item,
         selectedLoan: request.loan,
         selectedUser: request.requester,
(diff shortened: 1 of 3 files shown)


src/RequestForm.js
```diff
@@ -318,55 +312,61 @@ class RequestForm extends React.Component {
...
+        requestPreferencesLoaded: true,
+        hasDelivery,
+        defaultServicePointId,
+        defaultDeliveryAddressTypeId,
+      };
+
+      this.setState({ ...requestPreferences }, () => {
+        this.props.change('fulfilmentPreference', fulfillmentPreference);
+
+        if (isDelivery(this.props.initialValues)) {
```
MaksymDryha: I guess it should be `if (fulfillmentPreference === fulfilmentTypeMap.DELIVERY)`

src/RequestForm.js
```diff
@@ -318,55 +312,61 @@ class RequestForm extends React.Component {
...
+      const requestPreferences = {
+        ...this.getDefaultRequestPreferences(),
+        requestPreferencesLoaded: true,
+        hasDelivery,
+        defaultServicePointId,
+        defaultDeliveryAddressTypeId,
+      };
+
+      this.setState({ ...requestPreferences }, () => {
+        this.props.change('fulfilmentPreference', fulfillmentPreference);
```
MaksymDryha: Please, move `change` to destructuring assignment

src/RequestForm.js
```diff
@@ -318,55 +312,61 @@ class RequestForm extends React.Component {
...
+  async findRequestPreferences(userId) {
+    try {
+      const {
+        requestPreferences: [{
+          delivery: hasDelivery,
+          defaultServicePointId,
+          defaultDeliveryAddressTypeId,
+          fulfillment: fulfillmentPreference,
+        }]
+      } = await this.props.findResource('requestPreferences', userId, 'userId');
```
MaksymDryha: Please, move `findResource` to destructuring assignment

src/RequestForm.js
```diff
@@ -247,25 +243,13 @@ class RequestForm extends React.Component {
...
-    const { change } = this.props;
-
-    const deliverySelected = e.target.value === fulfilmentTypeMap.DELIVERY;
+    const selectedFullfillmentPreference = e.target.value;
 
     this.setState({
-      selectedDelivery: deliverySelected,
+      selectedDelivery: this.isDeliverySelected(selectedFullfillmentPreference),
+    }, () => {
+      this.updateRequestPreferencesFields(selectedFullfillmentPreference);
```
MaksymDryha: There's no need to pass `selectedFullfillmentPreference` to `updateRequestPreferencesFields` since all the needed data is stored in `state` which can be accessed from `updateRequestPreferencesFields`

src/RequestForm.js
```diff
@@ -318,55 +302,87 @@ class RequestForm extends React.Component {
...
+  }
+
+  updateRequestPreferencesFields(fullfillmentPreference) {
+    const {
+      defaultDeliveryAddressTypeId,
+      defaultServicePointId,
+    } = this.state;
+    const { change } = this.props;
+
+    if (fullfillmentPreference === fulfilmentTypeMap.DELIVERY) {
```
MaksymDryha: `state.selectedDelivery` can be used here making `fullfillmentPreference` parameter unnecessary

src/RequestForm.js
```diff
@@ -517,7 +533,7 @@ class RequestForm extends React.Component {
       pickupServicePointId,
     } = data;
 
-    const [block] = this.getPatronBlocks(parentResources);
+    const [block = {}] = this.getPatronBlocks(parentResources);
```
MaksymDryha: I think it's better to remove the default value here and then make the condition look like:
```
if (block && block.userId === this.state.selectedUser.id) 
```
or
```
if (get(block, 'userId ') === this.state.selectedUser.id) 
```

src/RequestForm.js
```diff
@@ -247,25 +243,13 @@ class RequestForm extends React.Component {
...
-      defaultServicePointId,
-    } = this.state;
-    const { change } = this.props;
-
-    const deliverySelected = e.target.value === fulfilmentTypeMap.DELIVERY;
+    const selectedFullfillmentPreference = e.target.value;
 
     this.setState({
-      selectedDelivery: deliverySelected,
+      selectedDelivery: this.isDeliverySelected(selectedFullfillmentPreference),
```
Yurii-Danylenko: Please, rename `selectedDelivery` everywhere in file to §deliveryIsSelected` or `deliverySelected`.

PR #468, opened 2019-12-12
UIREQ-370 - Create/Edit a request | Move Save/Cancel buttons to the footer

maximdidenkoepam:
# Purpose
Move navigation buttons to to footer.

# Link
https://issues.folio.org/browse/UIREQ-370

# Dependency
https://github.com/folio-org/platform-core/pull/491

# Screenshot
Screenshot 2019-12-12 at 16 01 01

### DIFF
--- a/src/RequestForm.js
+++ b/src/RequestForm.js
@@ -36,6 +36,7 @@ import {
   Row,
   Select,
   TextField,
+  PaneFooter,
 } from '@folio/stripes/components';
 
 import stripesForm from '@folio/stripes/form';
@@ -652,11 +653,13 @@ class RequestForm extends React.Component {
       },
       patronGroups,
       parentResources,
+      pristine,
       submitting,
       intl: {
         formatMessage,
       },
       errorMessage,
+      onCancel,
     } = this.props;
 
     const {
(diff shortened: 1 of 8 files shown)


YevheniiMaltsev:
@maximdidenkoepam Could you please add the information about changes to CHANGELOG.md?

translations/ui-requests/en.json
```diff
@@ -159,5 +155,8 @@
   "requestQueue.requests.notFound": "No requests found",
   "cancellationReason": "Cancellation reason",
   "cancellationAdditionalInformation": "Additional information for patron",
-  "printPickSlips": "Print pick slips for {sp}"
+  "printPickSlips": "Print pick slips for {sp}",
+
+  "common.cancel": "Cancel",
```
VictorSoroka1: I see that such translations exists in stripes-components and wonder whether it would be better to get them from there rather than introduce new ones here.

maximdidenkoepam: @VictorSoroka1, I prefer to have own translation in each module and not rely on translation from stripes-components, because in this case no one will know that this translation is used in another module and can change it in scope of some user story.

src/requests.css
```diff
@@ -11,5 +11,13 @@
...
   height: 100%;
+  width: 100%;
   overflow: auto;
 }
+
+.footerContent {
+  max-width: 50em;
+  display: flex;
+  justify-content: space-between;
+  width: 100%;
```
VictorSoroka1: Is the width really needed? Since the block is flex now, I think the container will fit 100% just because of that.

src/RequestForm.js
```diff
@@ -719,283 +633,301 @@ class RequestForm extends React.Component {
...
+                                <Field
+                                  name="item.barcode"
+                                  placeholder={placeholder}
+                                  aria-label={<FormattedMessage id="ui-requests.item.barcode" />}
+                                  fullWidth
+                                  component={TextField}
+                                  withRef
+                                  ref={this.itemBarcodeRef}
+                                  onKeyDown={e => this.onKeyDown(e, 'item')}
+
```
VictorSoroka1: Please remove the excess line.

src/RequestForm.js
```diff
@@ -719,283 +633,301 @@ class RequestForm extends React.Component {
...
+                      <Col xs={4}>
+                        {!isEditForm && !selectedItem &&
+                          <span data-test-request-type-message>
+                            <KeyValue
+                              label={<FormattedMessage id="ui-requests.requestType" />}
+                              value={<FormattedMessage id="ui-requests.requestType.message" />}
+                            />
+                          </span>
+                        }
+
```
VictorSoroka1: Please remove the excess line.

src/RequestForm.js
```diff
@@ -719,283 +633,301 @@ class RequestForm extends React.Component {
...
+                              marginTop0
+                              searchButtonStyle="link"
+                              {...this.props}
+                              dataKey="users"
+                              selectUser={this.onSelectUser}
+                              disableRecordCreation={disableRecordCreation}
+                              visibleColumns={['active', 'name', 'patronGroup', 'username', 'barcode']}
+                              columnMapping={columnMapping}
                             />
+
```
VictorSoroka1: Please remove the excess line.

PR #470, opened 2019-12-13
UIREQ-380: Fix displaying of delivery address

Vladyslav-Velytskyi:
# Purpose
Part of UIREQ-380. Fix bug of displaying delivery address when Fullfilment Preference is equal to Delivery in create/edit mode

# Link
https://issues.folio.org/browse/UIREQ-385

# Screenshots

### Create
oNfeA1DbxI

### Edit
OAYSuWjisk

### DIFF
--- a/src/RequestForm.js
+++ b/src/RequestForm.js
@@ -159,6 +159,10 @@ class RequestForm extends React.Component {
     if (this.props.query.itemId) {
       this.findItem('id', this.props.query.itemId);
     }
+
+    if (this.isEditForm()) {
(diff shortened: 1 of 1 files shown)


src/RequestForm.js
```diff
@@ -384,6 +408,10 @@ class RequestForm extends React.Component {
     return fulfillmentPreference === fulfilmentTypeMap.DELIVERY;
   }
 
+  setSelectedAddressTypeId(deliverySelected, defaultDeliveryAddressTypeId) {
```
MaksymDryha: Should it be `getSelectedAddressTypeId`?

PR #476, opened 2019-12-18
UIREQ-343: Fix open request count on item when creating a new request

mkuklis:
https://issues.folio.org/browse/UIREQ-343


src/routes/RequestsRoute.js
```diff
@@ -454,13 +454,11 @@ class RequestsRoute extends React.Component {
         this.findResource('user', r.requesterId),
         this.findResource('requestsForItem', r.itemId),
       ],
-    ).then((resultArray) => {
+    ).then(([users, requests]) => {
```
zburke: So much clearer; thank you!

PR #477, opened 2019-12-18
UIREQ-349: Fix links back to item record in new request form

Baroquem:
https://issues.folio.org/browse/UIREQ-349

### DIFF
--- a/src/RequestForm.js
+++ b/src/RequestForm.js
@@ -418,7 +418,7 @@ class RequestForm extends React.Component {
     return deliverySelected ? defaultDeliveryAddressTypeId : '';
   }
 
-  findLoan(item) {
+  findRelatedResources(item) {
     const { findResource } = this.props;
     if (!item) return null;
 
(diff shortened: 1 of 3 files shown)


src/RequestForm.js
```diff
@@ -418,20 +418,23 @@ class RequestForm extends React.Component {
...
     return Promise.all(
       [
         findResource('loan', item.id),
         findResource('requestsForItem', item.id),
+        findResource('holding', item.holdingsRecordId),
       ],
     ).then((results) => {
       const selectedLoan = results[0].loans[0];
       const requestCount = results[1].requests.length;
+      const holdingsRecord = results[2].holdingsRecords[0];
```
zburke:
```
.then(([loans, requests, holdingsRecords]) => {
  const selectedLoan = loans[0];
  ...
```
would be a bit clearer. I know you're working with existing code and this isn't a critique. I just happened to see a similar refactor in an unrelated PR today and thought, "ooooh, that's a nice improvement they snuck in there..."

Baroquem: I like the idea ... but because of the way the return values are structured here, this would actually require `const selectedLoan = loans.loans[0]`, which is no better.

PR #497, opened 2020-01-28
UIREQ-409 - added alphabetical order for list of available tags

maximdidenkoepam:
# Purpose
sort available tags in alphabetical order

# Link
https://issues.folio.org/browse/UIREQ-409

# Screenshot
Screen Shot 2020-01-28 at 12 06 22

### DIFF
--- a/src/components/RequestsFilters/RequestsFilters.js
+++ b/src/components/RequestsFilters/RequestsFilters.js
@@ -4,8 +4,6 @@ import { FormattedMessage } from 'react-intl';
 import {
   get,
   isEmpty,
-  chain,
-  map,
   sortBy,
 } from 'lodash';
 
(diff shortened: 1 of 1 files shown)


src/components/RequestsFilters/RequestsFilters.js
```diff
@@ -37,10 +40,12 @@ export default class RequestsFilters extends React.Component {
   };
 
   transformTagsOptions = () => {
-    return get(this.props.resources, 'tags.records', []).map(({ label }) => ({
-      label,
-      value: label,
-    }));
+    const tags = get(this.props.resources, 'tags.records', []);
+
+    return chain(tags)
```
VictorSoroka1: It would be great not to use chain, as it actually the wrapper on the whole lodash.
`_.chain = (array) => wrap(array, _); // Rough concept of chain`

So the `map` which is imported above is not actually used. Instead used the `.map` provided by a chain. Please read more information about it in the [article](https://medium.com/making-internets/why-using-chain-is-a-mistake-9bc1f80d51ba).

PR #503, opened 2020-02-05
UIREQ-370 backport "Move Save/Cancel buttons to the footer" to v1.14

zburke:
Backport the PRs related to UIREQ-370, Move Save/Cancel buttons to the footer, to the 2019-q4 branch (`b1.14-patch`) in preparation for a patch release. 

Refs [UIREQ-370](https://issues.folio.org/browse/UIREQ-370)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for ui-requests
 
+## 1.14.4 (IN PROGRESS)
+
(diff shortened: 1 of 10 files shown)


zburke:
These unit tests pass when this module is built against a released version of stripes. 

Here in CI, they are failing because they are compiled against the stripes-components master branch which has some incompatible HTML markup. The API itself has not changed (hence no major version change) but the internal representation the unit tests reach for is different.

PR #508, opened 2020-02-07
Revert "UIREQ-418 update tests for action menu"

zburke:
Reverts folio-org/ui-requests#507. #501 already did all of this and more.


zburke:
In fact, @annamelnyk has already updated #501 to accommodate the test-changes in #507, so we'll leave it in place.

PR #514, opened 2020-02-14
UIREQ-395: Make Export hold shelf clearance report link disabled

annamelnyk:
# Description
 Make `Export hold shelf clearance report` link for `CurrentServicePoint` disabled if there is nothing on it for the currently selected service point.
#Approach
We are looking if `user.curServicePoint.id` exists:

- if no:  modal error is shown

- if yes: we check if there are any `totalRecords `after searching and filtering results, and if there any records in the current `servicePoint`

# Issue
https://issues.folio.org/browse/UIREQ-395
# Screenshots
image
image


test/bigtest/tests/export-to-CSV-test.js
```diff
@@ -80,35 +80,10 @@ describe('Export to CSV', () => {
...
       beforeEach(async function () {
         await requests.headerDropdown.click();
-        await requests.headerDropdownMenu.clickExportExpiredHoldsToCSV();
       });
 
       it('exports data to csv', () => {
-        expect(requests.headerDropdownMenu.exportExpiredHoldsBtnIsVisible).to.be.false;
-      });
-
-      describe('error modal ', () => {
```
VictorSoroka1: Could you please clarify the reason for removing tests for error modal?

src/routes/RequestsRoute.js
```diff
@@ -365,6 +368,7 @@ class RequestsRoute extends React.Component {
     const { mutator } = this.props;
     const { id } = this.getCurrentServicePointInfo();
     mutator.currentServicePoint.update({ id });
+    this.buildRecordsForHoldsShelfReport();
```
maximdidenkoepam: I think generate report each time when component re-renders not a good idea. If it possible to find a condition when it shout be done?

test/bigtest/tests/export-to-CSV-test.js
annamelnyk: This modal was removed from component and from tests as well, because of the **Export hold shelf clearance report** should be disabled if there is nothing on this report for the currently selected service point, we do not need this error modal any more, and also it will never show because we cannot click to disabled link

PR #524, opened 2020-03-05
UIREQ-372: Display effective call number string on the request queue …

SergiyVSergiyenko:
## Purpose
Display effective call number string on the dedicated request queue page. [Story](https://issues.folio.org/browse/UIREQ-372).

### Screenshots
**Item data**
Item_data

**Call number string**
Call_number_string

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,7 +143,7 @@
     "@bigtest/mocha": "^0.5.2",
     "@bigtest/react": "^0.1.2",
     "@folio/eslint-config-stripes": "^5.1.0",
-    "@folio/stripes": "^2.13.0",
+    "@folio/stripes": "^3.0.0",
     "@folio/stripes-cli": "^1.11.0",
     "@folio/stripes-core": "^3.12.0",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -160,6 +160,7 @@
   },
   "dependencies": {
     "@folio/react-intl-safe-html": "^1.0.2",
+    "@folio/stripes-util": "^2.0.0",
```
zburke: Please import this via `@folio/stripes/util` instead.

zburke: i.e. we don't need a direct dependency because we can rely on the `@folio/stripes` peerDependency.

SergiyVSergiyenko: @zburke, I don't know why, but when importing via '@folio/stripes/util' the 'copyNumber' value doesn't appear in the 'Effective call number string' field.

zburke: What the ...? Makes me wonder if the correct version of stripes-util is being wrapped into stripes.

PR #527, opened 2020-03-11
UIREQ-372-v2: Display effective call number string on the dedicated r…

SergiyVSergiyenko:
## Purpose
Display effective call number string on the dedicated request queue page. [Story](https://issues.folio.org/browse/UIREQ-372).


SergiyVSergiyenko:
I'm going to close the previous PR https://github.com/folio-org/ui-requests/pull/524 due to the incomprehensible behavior of dependencies.

zburke:
Ah; we're typing at the same time, and arriving at the same conclusions. Good!

PR #532, opened 2020-03-16
UIREQ-417 - New/Edit Request: Cancel and "X" does not work as expected once you click Keep editing

maximdidenkoepam:
# Purpose
Allow user cancel new request after сlick on keep editing on previous cancelation.

# Link
https://issues.folio.org/browse/UIREQ-417

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,9 +1,12 @@
 import {
   get,
+  forOwn,
+  isEmpty,
   isObject,
   omit,
   cloneDeep,
 } from 'lodash';
+import queryString from 'query-string';
 import React from 'react';
 import { Link } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -208,3 +211,21 @@ export const convertToSlipData = (source, intl, timeZone, locale, slipName = 'Pi
     };
   });
 };
+
+export function buildUrl(location, values) {
+  let url = values._path || location.pathname;
+  const locationQuery = location.query ? location.query : queryString.parse(location.search);
+  const params = {};
+
+  forOwn(Object.assign(locationQuery, values), (value, key) => {
```
mkuklis: Could we use `omit` here instead? https://lodash.com/docs/4.17.15#omit

maximdidenkoepam: Done

PR #534, opened 2020-03-17
upgrade plugin-find-user to v2.0.0

zburke:
Update the optional-dep on plugin-find-user to one that uses a
compatible version of `@folio/stripes`.


zburke:
Refs [UIREQ-450](https://issues.folio.org/browse/UIREQ-450)

PR #538, opened 2020-03-27
CHAL-91 show "Load more" button

zburke:
Replace infinite scroll with a "Load more" button in the results list.
Infinite scroll can behave poorly when results lists are long.

Refs [CHAL-91](https://issues.folio.org/browse/CHAL-91)


zburke:
Never got this to run cleanly in CI even though tests run just fine locally. Shame.

PR #546, opened 2020-04-15
UIREQ-444: Add ability to create a request with the requester without barcode

mkuklis:
### Purpose
Add ability to create a request with the requester without barcode.

### Link
https://issues.folio.org/browse/UIREQ-444

request-form
request_view


SergiyVSergiyenko:
Great tests! 👏  Thank you for showing how it is possible to connect modules in the 'set-application'.

PR #557, opened 2020-04-22
UIREQ-437: Remove duplicated ids

annamelnyk:
# Description
Fix duplicated ids: `clickable-cancel-request` id was already existed on `Edit` page
# Story
In scope of https://issues.folio.org/browse/UIREQ-437


annamelnyk:
Thanks for comment @zburke, I totally agree that we should work on semantically proper and relevant names  for attributes to be appropriate with what current element is responsible for

PR #566, opened 2020-04-30
UIREQ-456: Omit proxyUserId from duplicated properties

Baroquem:
https://issues.folio.org/browse/UIREQ-456

Copying the proxyUserId when duplicating a request that has a proxy user causes an error when trying to save the new request.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -107,6 +107,7 @@ export function duplicateRequest(request) {
     'requester',
     'item',
     'pickupServicePoint',
+    'proxyUserId'
   ]);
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -107,6 +107,7 @@ export function duplicateRequest(request) {
     'requester',
     'item',
     'pickupServicePoint',
+    'proxyUserId'
```
SergiyVSergiyenko, line 110:
```suggestion
    'proxyUserId',
```

PR #570, opened 2020-05-07
UIREQ-390: Prevent requests for Withdrawn items

SergiyVSergiyenko:
## Purpose
Prevent requests for items with the status `Withdrawn`. Story [UIREQ-390](https://issues.folio.org/browse/UIREQ-390).

### Approach
Since the prevention of requests for items with the status `Declared lost` and `Withdrawn` is performed in a similar way, the code was refactored to use one functionality in both cases.

### Screenshot
Screen Shot 2020-05-07 at 15 30 36

### DIFF
--- a/src/RequestForm.js
+++ b/src/RequestForm.js
@@ -65,6 +65,7 @@ import {
   getRequestTypeOptions,
   isDelivery,
   isDeclaredLostItem,
+  isWithdrawnItem,
 } from './utils';
 
 import css from './requests.css';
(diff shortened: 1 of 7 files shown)


src/RequestForm.js
```diff
@@ -1046,19 +1051,19 @@ class RequestForm extends React.Component {
               viewUserPath={() => this.onViewUserPath(selectedUser, patronGroup)}
               patronBlocks={patronBlocks || []}
             />
-            {showDeclaredLostError &&
+            {isErrorModalOpen &&
               <ErrorModal
-                id="declared-lost-modal"
-                data-test-declared-lost-modal
```
SergiyVSergiyenko, line 1052: `data-test` attribute has been removed due to not being used anywhere (at least for now). For tests, the `data-test-error-modal` attribute built into ErrorModal is used.

PR #582, opened 2020-06-05
UIREQ-471 - Patron Blocks: Is there an automated patron block on requesting?

maximdidenkoepam:
# Purpose
Add automated patron blocks support for requests.

# Link
https://issues.folio.org/browse/UIREQ-471

# Screenshot
Screen Shot 2020-06-05 at 17 55 30

### DIFF
--- a/src/routes/RequestsRoute.js
+++ b/src/routes/RequestsRoute.js
@@ -150,6 +150,11 @@ class RequestsRoute extends React.Component {
         path: 'manualblocks/%{activeRecord.blockId}',
       },
     },
+    automatedPatronBlocks: {
+      type: 'okapi',
+      records: 'automatedPatronBlocks',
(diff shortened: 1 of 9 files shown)


src/routes/RequestsRoute.js
```diff
@@ -150,6 +150,11 @@ class RequestsRoute extends React.Component {
         path: 'manualblocks/%{activeRecord.blockId}',
       },
     },
+    automatedPatronBlocks: {
+      type: 'okapi',
+      records: 'automatedPatronBlocks',
+      path: 'automated-patron-blocks/%{activeRecord.patronId}',
+    },
```
annamelnyk, line 157: One patron can have above 100 automated blocks. Probably `limit` should be added here

maximdidenkoepam: The story has no requirements about restrictions, I think we will be able to add limit later in case of necessarily.

PR #590, opened 2020-06-17
UIREQ-457: Add staff notes accordion to request details page

Vladyslav-Velytskyi:
## Description

- Add `Staff notes` accordion to request details pane
- Add `New staff notes` page

## Issue
[UIREQ-457](https://issues.folio.org/browse/UIREQ-457)

### DIFF
--- a/src/routes/NoteCreateRoute.js
+++ b/src/routes/NoteCreateRoute.js
@@ -1,0 +1,40 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ReactRouterPropTypes from 'react-router-prop-types';
+import { Redirect } from 'react-router';
+import { FormattedMessage } from 'react-intl';
+
+import { NoteCreatePage } from '@folio/stripes/smart-components';
+
+import {
+  DOMAIN_NAME,
+  APP_ICON_NAME,
+} from '../constants';
+import { formatNoteReferrerEntityData } from '../utils';
+
+const NoteCreateRoute = ({
+  location,
+  history,
+}) => {
+  return location.state
+    ? (
+      <NoteCreatePage
+        referredEntityData={formatNoteReferrerEntityData(location.state)}
+        entityTypeTranslationKeys={{ request: 'ui-requests.notes.entityType.request' }}
+        paneHeaderAppIcon={APP_ICON_NAME}
+        paneTitle={<FormattedMessage id="ui-request.notes.newStaffNote" />}
+        domain={DOMAIN_NAME}
+        navigateBack={history.goBack}
+      />
+    )
+    : <Redirect to="/requests" />;
+};
+
+NoteCreateRoute.propTypes = {
+  history: PropTypes.shape({
+    goBack: PropTypes.func.isRequired,
+  }).isRequired,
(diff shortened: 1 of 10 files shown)


test/bigtest/tests/view-request-test.js
```diff
@@ -86,6 +86,20 @@ describe('View request page', () => {
...
+    it('should display staff notes accordion', () => {
+      expect(viewRequest.staffNotesAccordion.isPresent).to.be.true;
+    });
+
+    describe('when click new button in staff notes accordion', () => {
+      beforeEach(async () => {
+        await viewRequest.staffNotesAccordion.clickNewButton();
+      });
+
+      it('should displayed new staff notes page', function () {
```
VictorSoroka1, line 99:
```suggestion
      it('should display new staff notes page', function () {
```

test/bigtest/tests/utils-test.js
```diff
@@ -172,4 +173,25 @@ describe('utils', () => {
...
+      const expectedData = {
+        name: data.entityName,
+        type: data.entityType,
+        id: data.entityId,
+      };
+
+      expect(formatNoteReferrerEntityData(data)).to.deep.equal(expectedData);
+    });
+
+    it('should return false', () => {
```
VictorSoroka1, line 193: Could you please extend the description for this case so we reason about it without looking at the actual code.
```suggestion
    it('should return false', () => {
```

src/utils.js
```diff
@@ -238,3 +238,21 @@ export function buildUrl(location, values) {
 
   return isEmpty(params) ? url : `${url}?${queryString.stringify(params)}`;
 }
+
+export function formatNoteReferrerEntityData(data) {
+  if (data) {
```
VictorSoroka1, line 243: I would suggest checking firstly for `!data` so we don't have the code indented when `data` is truthy.

src/routes/NoteCreateRoute.js
```diff
@@ -0,0 +1,40 @@
...
+}) => {
+  return location.state
+    ? (
+      <NoteCreatePage
+        referredEntityData={formatNoteReferrerEntityData(location.state)}
+        entityTypeTranslationKeys={{ request: 'ui-requests.notes.entityType.request' }}
+        paneHeaderAppIcon={APP_ICON_NAME}
+        paneTitle={<FormattedMessage id="ui-request.notes.newStaffNote" />}
+        domain={DOMAIN_NAME}
+        navigateBack={history.goBack}
```
VictorSoroka1, line 27: Could you please clarify how this callback eventually used? I have concern over the case when the user opened the page on the route (by URL in the new tab let's say) from where then the user is doing the navigation back without previous history. In this case, the user will be redirected out of the folio app.

Vladyslav-Velytskyi: It works like a simple browser back. In case that you described, user will redirect to ``/request``, because `location.state` should be empty. Also this component was implemented by identity the same one in `ui-eholdings` module - [link](https://github.com/folio-org/ui-eholdings/blob/master/src/routes/note-create.js)

VictorSoroka1: Sounds good.

PR #602, opened 2020-06-30
UIREQ-481: Fix ability to sort by Request Date column

SergiyVSergiyenko:
https://issues.folio.org/browse/UIREQ-481

### Purpose
The value for sorting was taken from the column labels, which led to a dependence on the tenant's locale.
This PR fixed this by providing the changes that ensure that the sort value refers to the column name.

### DIFF
--- a/src/routes/RequestsRoute.js
+++ b/src/routes/RequestsRoute.js
@@ -81,7 +81,7 @@ class RequestsRoute extends React.Component {
       records: 'addressTypes',
     },
     query: {
-      initialValue: { sort: 'Request date' },
+      initialValue: { sort: 'requestDate' },
     },
     resultCount: { initialValue: INITIAL_RESULT_COUNT },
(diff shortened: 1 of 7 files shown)


src/routes/RequestsRoute.js
```diff
@@ -102,8 +102,8 @@ class RequestsRoute extends React.Component {
               'Item barcode': 'item.barcode',
               'Type': 'requestType',
               'Requester': 'requester.lastName requester.firstName',
-              'Requester barcode': 'requester.barcode',
-              'Request date': 'requestDate',
+              'Requester Barcode': 'requester.barcode',
+              'Request Date': 'requestDate',
               'Position': 'position',
               'Proxy': 'proxy',
```
zburke, line 108: TBH, I'm surprised these don't need to match the keys specified in `this.columnLabels` starting on Line 282. Does sorting still work on this branch if your local is something other than `en-US`?

PR #605, opened 2020-07-02
CIRC-817: Use == for requestType filters

mkuklis:
This should improve the performance a bit when filtering by request types.

https://issues.folio.org/browse/CIRC-807


mkuklis:
Thank you for catching this @zburke. I just added it to statuses.

PR #620, opened 2020-08-04
UIREQ-508: Add loading indicator when service point is switched

mkuklis:
### Purpose

Add loading indicator when service point is switched and the pick slips are being reloaded.

### Link
https://issues.folio.org/browse/UIREQ-508

### Screenshot
slips_loading

### DIFF
--- a/src/routes/RequestsRoute.js
+++ b/src/routes/RequestsRoute.js
@@ -739,7 +739,7 @@ class RequestsRoute extends React.Component {
           <FormattedMessage id="ui-requests.exportSearchResultsToCsv" />
         </Button>
         {
-          !pickSlipsArePending ?
+          pickSlipsArePending ?
             <LoadingButton>
               <FormattedMessage id="ui-requests.pickSlipsLoading" />
             </LoadingButton> :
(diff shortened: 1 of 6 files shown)


src/routes/RequestsRoute.js
```diff
@@ -728,34 +738,42 @@ class RequestsRoute extends React.Component {
...
+                template={this.getPrintTemplate()}
+                dataSource={convertToSlipData(pickSlips, intl, timezone, locale)}
+                onBeforePrint={onToggle}
+              >
+                <FormattedMessage
+                  id="ui-requests.printPickSlips"
+                  values={{ sp: servicePointName }}
+                />
+              </PrintButton>
+            </> //
```
Baroquem, line 774: Are there comments on these lines? I'm not sure what the // are there for.

mkuklis: could catch @Baroquem. I added them there because my editor gets confused with `</>` and screws up the color theme. I just cleaned that up.

mkuklis:
I haven't looked at UIREQ-507 yet but I noticed that couple times my chrome browser got completely frozen when I switched between service points (even before clicking on print). I worry there could be something else going on here.

PR #626, opened 2020-08-20
UIREQ-508: Add feedback after print button is clicked

mkuklis:
For a very large number of slips it was not clear if the print process has started or not. This PR adds a callout and closes the menu after the print button is clicked.

https://issues.folio.org/browse/UIREQ-508

printing


SergiyVSergiyenko (approved):

SergiyVSergiyenko:
Pretty witty solution, @mkuklis! 👍

PR #629, opened 2020-08-24
UIREQ-429: Block requests for Aged to lost items

Baroquem:
https://issues.folio.org/browse/UIREQ-429

This is basically just adding 'Aged to lost' to the existing code that throws up an error modal for various item statuses for which requests are disallowed.

Screen Shot 2020-08-24 at 4 50 07 PM


SergiyVSergiyenko (approved):

SergiyVSergiyenko:
It looks pretty simple! 😄  But we all know how much has been done to achieve this ease. Well done, @Baroquem! 👍

PR #634, opened 2020-09-11
STRIPES-693 Apply columnwidths API to MCL

JohnC-80:
In the spirit of sparing *some horizontal space... This almost knocked off a whole column width from the example data...

## Before
Requests-before

## After
Requests-updated


mkuklis (approved):
Oh this is so much better @JohnC-80. Nice work!

PR #643, opened 2020-10-05
UIREQ-514: Select requester validation error needs rewording

K-Felk:
I reworded the string in the master translation file.

refs: https://issues.folio.org/browse/UIREQ-514


Baroquem (approved):

Baroquem (commented):
Oops -- going by the quoted message in the issue, we should use the spelling "requester" rather than "requestor".

PR #662, opened 2020-11-09
UIREQ-539: Fix bug limiting number of records in CSV search results export

Baroquem:
https://issues.folio.org/browse/UIREQ-539

The CSV search results export was relying on the same resource that was used to display (paginated) records in the UI. To provide a comprehensive export of all records matching the search and filter criteria requires being able to retrieve tens of thousands of records (the Honeysuckle bugfest environment contains almost 30,000 request records). 

The solution I've come up with here is not very pretty, and there may well be a better approach that I'm overlooking. I would welcome any suggestions for alternatives!


zburke (approved):
I'm not super stoked about duplicating the query string from the `records` resource inside `exportData`; could that be easily separated out into a local function (outside the class, similar to how `urls` is initialized)? Likewise, it feels like `buildFilterQuery` is duplicating logic we must have elsewhere (is it part of `SearchAndSort`?).

If trying to reuse all that logic is too cumbersome -- and it may be, that's fine -- can you at least stick some comments around the pieces that should be kept in sync?

Baroquem:
> I'm not super stoked about duplicating the query string from the `records` resource inside `exportData`; could that be easily separated out into a local function (outside the class, similar to how `urls` is initialized)? Likewise, it feels like `buildFilterQuery` is duplicating logic we must have elsewhere (is it part of `SearchAndSort`?).
> 
> If trying to reuse all that logic is too cumbersome -- and it may be, that's fine -- can you at least stick some comments around the pieces that should be kept in sync?

@zburke These are good ideas. My instincts are similar, but I ran into problems trying to execute them. The duplicated query string came about because the query substitution for `%{query.query}` simply would not work for some reason, and I had to resort to the `${queryTerm}` form.

Similarly, I hunted around for an established function for building a filter query, but didn't find anything that quite did the trick. It's entirely possible that I overlooked something, though. Your questions have inspired me to go back and do some more searching.

Baroquem:
Just pushed a slightly cleaner version.

mkuklis (approved):
Thank you for spending extra time with this @Baroquem!

zburke (approved):
Nice work!

PR #669, opened 2020-11-19
await request details pane

zburke:
Failing tests in #668 appear to be timing related, failing to await the
display of the request-details pane (or running out of time before it
displays). Wish I knew why this started showing up all of a sudden, but
here we are.

### DIFF
--- a/test/bigtest/tests/new-request-test.js
+++ b/test/bigtest/tests/new-request-test.js
@@ -26,7 +26,7 @@ const itemStatuses = [
   'Aged to lost',
 ];
 
-describe('New Request page', () => {
+describe.only('New Request page', () => {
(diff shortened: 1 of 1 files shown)


test/bigtest/tests/new-request-test.js
```diff
@@ -26,7 +26,7 @@ const itemStatuses = [
   'Aged to lost',
 ];
 
-describe('New Request page', () => {
+describe.only('New Request page', () => {
```
Baroquem, line 29: You've got a rogue `only` here.

SergiyVSergiyenko (approved):

mkuklis (approved):
Looks good after the `only` is removed.

PR #672, opened 2020-11-20
UIREQ-558: Retrieve requests to items in chunks

SergiyVSergiyenko:
https://issues.folio.org/browse/UIREQ-558

The problem for which this PR is addressed has the same origin as the problem described in [#974](https://github.com/folio-org/ui-users/pull/974). And the solution is the same.
The only difference is that chunks of 40 are used to retrieve open requests against items.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -263,9 +263,3 @@ export function formatNoteReferrerEntityData(data) {
     id,
   };
 }
-
-export function eachPromise(arr, fn) {
-  if (!Array.isArray(arr)) return Promise.reject(new Error('Array not found'));
-
-  return arr.reduce((acc, cur) => acc.then(response => fn(response, cur)), Promise.resolve([]));
-}
--- a/src/utils.js
+++ b/src/utils.js
@@ -263,3 +263,9 @@ export function formatNoteReferrerEntityData(data) {
     id,
   };
 }
+
+export function eachPromise(arr, fn) {
+  if (!Array.isArray(arr)) return Promise.reject(new Error('Array not found'));
+
+  return arr.reduce((acc, cur) => acc.then(response => fn(response, cur)), Promise.resolve([]));
+}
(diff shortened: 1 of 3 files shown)


Baroquem (approved):

mkuklis (approved):
@SergiyVSergiyenko this looks good. Do you think that executing them all at once instead of in the sequence will work fine here? 

I'm asking because I've seen the backend not being able to handle more than couple requests at the same time in the past.

zburke (approved):
Good point, @mkuklis. See https://github.com/folio-org/stripes-util/blob/master/lib/eachPromise.js for an alternative implementation.

src/utils.js
```diff
@@ -263,3 +263,9 @@ export function formatNoteReferrerEntityData(data) {
     id,
   };
 }
+
+export function eachPromise(arr, fn) {
+  if (!Array.isArray(arr)) return Promise.reject(new Error('Array not found'));
+
+  return arr.reduce((acc, cur) => acc.then(response => fn(response, cur)), Promise.resolve([]));
+}
```
zburke, line 271: The point of using `reduce` rather than `Promise.all` is to cause the fetches to be processed in series rather than in parallel. You can see how it was [used for DELETE and POST](https://github.com/folio-org/ui-users/pull/1476/files) requests in ui-users; maybe we just need to configure the GET mutator correctly -- set `accumulate: true`, I think -- in order to make it return a Promise so we can use [the stripes-util version of `eachPromise`](https://github.com/folio-org/stripes-util/blob/6b8cf3772d82ba4c7a076c099563e00415899c0f/lib/eachPromise.js)?

mkuklis (approved):
Nice work @SergiyVSergiyenko!

Baroquem (approved):

zburke (approved):

PR #675, opened 2020-11-26
UIREQ-566: Increase limit to display correct number of requests in Mo…

SergiyVSergiyenko:
https://issues.folio.org/browse/UIREQ-566

### Purpose
Since on `bugfest-honeysuckle` we already have instances containing over 100 items and all / most of those items may have a request (or requests) for them, it was decided to increase the limit to display the correct number of requests for each item in the `Move request` modal.

### DIFF
--- a/src/MoveRequestDialog.js
+++ b/src/MoveRequestDialog.js
@@ -154,7 +154,7 @@ class MoveRequestDialog extends React.Component {
     query = `(${query}) and (status="Open")`;
     requests.reset();
 
-    return requests.GET({ params: { query } });
+    return requests.GET({ params: { query, limit: 1000 } });
   }
 
   async onRowClick(item) {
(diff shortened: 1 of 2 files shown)


ViktorSoroka07 (approved):

src/MoveRequestDialog.js
```diff
@@ -154,7 +154,7 @@ class MoveRequestDialog extends React.Component {
     query = `(${query}) and (status="Open")`;
     requests.reset();
 
-    return requests.GET({ params: { query } });
+    return requests.GET({ params: { query, limit: 1000 } });
```
zburke, line 157: Is it seriously the case that if you ask for N records by ID, e.g. `itemId==foo or itemId==bar`, that you only get ten records, not N records, unless you specify a `limit` clause? 🤦 

Also, we need #672 in place first in order to be able to fetch more than ~40 rows, right?

SergiyVSergiyenko: You know that without specifying a limit, we only get default = 10.

If the list of elements is split (40 for each chunk), there is a chance that we will also receive several hundred requests for each chunk.

PR #690, opened 2021-01-08
STRIPES-721: Replace withRef with forwardRef

mkuklis:
Due to the upgrade of redux and react-redux in https://issues.folio.org/browse/STRIPES-721 the `withRef` was removed and it has to be replaced with `forwardRef`.


zburke:
Do we need to update the `@folio/stripes` dep to v6, and hence the major version for this package as well? (But only after we actually merge all those [STRIPES-721](https://issues.folio.org/browse/STRIPES-721) PRs?

mkuklis:
Things still seem to work fine for me after this change but I think it's better to bump it here and wait. Good idea @zburke.

SergiyVSergiyenko (approved):

PR #710, opened 2021-02-23
UIREQ-543: Fix canceled request display in Hold shelf clearance report

SergiyVSergiyenko:
https://issues.folio.org/browse/UIREQ-543

### Purpose
Add the ability to get the canceled request data immediately after the request cancellation event (without refreshing the page) to build an up-to-date `Hold shelf clearance report`.

### DIFF
--- a/src/routes/RequestsRoute.js
+++ b/src/routes/RequestsRoute.js
@@ -355,7 +355,6 @@ class RequestsRoute extends React.Component {
 
   componentDidMount() {
     this.setCurrentServicePointId();
-    this.buildRecordsForHoldsShelfReport();
   }
 
   componentDidUpdate(prevProps) {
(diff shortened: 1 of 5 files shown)


src/routes/RequestsRoute.js
```diff
@@ -354,7 +354,6 @@ class RequestsRoute extends React.Component {
 
   componentDidMount() {
     this.setCurrentServicePointId();
-    this.buildRecordsForHoldsShelfReport();
```
SergiyVSergiyenko, line 357: The body of the `setCurrentServicePointId` function contains an explicit call to the `buildRecordsForHoldsShelfReport` function.

zburke (approved):
Checking my understanding: this basically puts a call to rebuild the report into the "cancel request" handler. I'm assuming that we build the report async because it's expensive to construct. Is there any concern that re-constructing it every time a request is canceled will also negatively impact performance? Maybe it's a legit tradeoff, I just want to be sure we're thinking about all the implications.

mkuklis (approved):

Baroquem (approved):

SergiyVSergiyenko:
Yes, @zburke, I can understand your concern. We create an additional request to the server every time when a request for an item is canceled, even in those cases when we are not going to export the corresponding report.
Another approach that might be here is to place a call to rebuild the report into the "export report" handler.
However, this requires changing the order in which the `Export hold shelf clearance report` button is displayed (it must be constantly active).

PR #711, opened 2021-02-24
UIREQ-576: Override patron blocks

annamelnyk:
# Description
Patron blocks: Allow for override of Requesting when logged in user has credentials
# Steps to reproduce
Need to create request for patron who has Patron block for Requesting
# Story
https://issues.folio.org/browse/UIREQ-576
# Screenshot
image


mkuklis (approved):

Dmitriy-Litvinenko (approved):

src/components/ErrorModal.js
```diff
@@ -1,6 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
+import {
+  map,
+  split,
+} from 'lodash';
```
zburke, line 7: OOC why are we importing this stuff from lodash? Are the native implementations insufficient for some reason?

zburke (approved):

Baroquem (approved):

src/components/ErrorModal.js
annamelnyk: @zburke, this is a good question. Actually I was thinking that we use `lodash` for performance improvements.

PR #730, opened 2021-04-14
UIREQ-601: Patron block modal does not show up

annamelnyk:
# Description
Patron block modal does not show up when we come to request page with predefined user again.
# Issue
https://issues.folio.org/browse/UIREQ-601


src/RequestForm.js
```diff
@@ -381,6 +381,11 @@ class RequestForm extends React.Component {
         if ((blocks.length && blocks[0].userId === selectedUser.id) || (!isEmpty(automatedPatronBlocks) && !isAutomatedPatronBlocksRequestInPendingState)) {
           state.blocked = true;
           state.isPatronBlocksOverridden = false;
+
+          this.setState({
```
mkuklis, line 385: @annamelnyk this looks good but I'm wondering do we still need line 382 and 383 or could we remove it?

````js
state.blocked = true;
state.isPatronBlocksOverridden = false;
````

mkuklis (approved):

src/RequestForm.js
annamelnyk: Thanks, removed it

Dmitriy-Litvinenko (approved):

PR #733, opened 2021-04-18
UIREQ-597: Change request type when changing to an item with different status

Baroquem:
https://issues.folio.org/browse/UIREQ-597

If a new request is created for an available item, the request type will be "page". If the requested item is then changed to one that is unavailable, the UI correctly switches to a drop-down with "hold" and "recall" options -- but the actual request type wasn't being changed, so an error would occur if the user didn't select one of the new options manually. We just need to ensure that the request type is changed to match what the UI displays.


mkuklis (approved):
Nice catch @Baroquem!

PR #746, opened 2021-05-18
UIREQ-607: Add translation key "mod-circulation.requesterHasItemOnLoan"

julianladisch:
mod-circulation may send the translation key

"ui-requests.mod-circulation.requesterHasItemOnLoan".

This PR only adds the translation key.

Processing the translation key is out of scope.
Generating the translation key is out of scope.

For details see

* https://issues.folio.org/browse/UIREQ-603 "Internationalize back-end failure messages"
* https://issues.folio.org/browse/CIRC-1146 "Internationalize back-end failure messages"
    = https://github.com/folio-org/mod-circulation/pull/879/files


zburke:
Belatedly closing this since [the TC has concluded translation of backend strings should happen in backend modules](https://wiki.folio.org/display/DD/Handle+i18n+where+messages+are+generated).

PR #802, opened 2021-11-23
TLR data migration

Dmitriy-Litvinenko:
## Purpose
TLR data migration

## Approach
This is feature branch for title level requests - data migration (https://issues.folio.org/browse/UXPROD-3330). We add all necessary changes. List of changes below:
- request list, view, create, edit, duplicate(UIREQ-664).
- dialogboxes and reorder list(UIREQ-665).
- hold shelf clearance report(UIREQ-677) .
- move `holdingsRecordId`(UIREQ-685).

## Refs
https://issues.folio.org/browse/UIREQ-664
https://issues.folio.org/browse/UIREQ-665
https://issues.folio.org/browse/UIREQ-677
https://issues.folio.org/browse/UIREQ-685


Dmitriy-Litvinenko:
Approved

Stanislau-Snapok (approved):

pavelspichonak (approved):

package.json
```diff
@@ -29,7 +29,7 @@
     "queryResource": "query",
     "okapiInterfaces": {
       "cancellation-reason-storage": "1.1",
-      "circulation": "10.0 11.0",
+      "circulation": "10.0 11.0 12.0",
```
zburke, line 32: Looking at the changes to the shape of the request data where `item` has been replaced by `instance`, do we really continue to support versions `10.0` and `11.0`? It feels like we should instead support `12.0` exclusively and bump the module's major version.

src/routes/RequestsRoute.js
```diff
@@ -115,9 +115,9 @@ class RequestsRoute extends React.Component {
         params: {
           query: makeQueryFunction(
             'cql.allRecords=1',
-            '(requesterId=="%{query.query}" or requester.barcode="%{query.query}*" or item.title="%{query.query}*" or item.barcode=="%{query.query}*" or itemId=="%{query.query}" or itemIsbn="%{query.query}")',
+            '(requesterId=="%{query.query}" or requester.barcode="%{query.query}*" or instance.title="%{query.query}*" or item.barcode=="%{query.query}*" or itemId=="%{query.query}" or itemIsbn="%{query.query}")',
```
zburke, line 118: Should `itemId=="%{query.query}"` remain here, or should it be replaced by/supplemented by `instanceId=="%{query.query}"`

zburke (changes requested):
1. It isn't clear to me that ILR and TLR requests will really co-exist after this work given that in most cases, `item.${field}` is simply replaced by `instance.${field}`. For an ILR, will these fields in fact be available at the "instance" level of the request?
2. Have all these instance-level fields really been available since `circulation` `10.0` and we're just now adding support for them in the UI? If that's the case, this is fine, but if these are only `12.0` features, we either need to change this implementation to accommodate requests that match the old shape or we need to drop support for the old interfaces.

src/routes/RequestsRoute.js
Stanislau-Snapok: `itemId` will be at the same place after migration. probably we need to add instanceId.

package.json
Stanislau-Snapok: Yeah, you right, I think we should use only 12 version. Because after migration previous versions will send invalid responce.

Stanislau-Snapok: Done. Could you please clarify, if we affect only `requests` module, should we leave last three versions in other modules, or only `12.0` as here?

src/routes/RequestsRoute.js
Stanislau-Snapok: Done. Added `instanceId` into query string.

package.json
zburke: It's up to you whether to continue supporting older interfaces in other modules. Here, we were clearly adding v12-specific code. If other modules that use the circulation interface don't do anything that is v12-specific, there is no reason to drop support for those version.

Stanislau-Snapok: Ok. I think we can just add `12.0` in other modules, because we have breaking changes only for `request` module.
Can we count this conversation as resolved?
And could you please approve this PR, if now everything looks good for you.
Thanks!

zburke (changes requested):
Removing support for an interface is a breaking change and requires incrementing the major version. Please update it in `package.json` and make noisy note of it in the readme with something like "BREAKING CHANGE: in adding TLRs we removed support for interfaces that only handle item-level requests."

Stanislau-Snapok:
> Removing support for an interface is a breaking change and requires incrementing the major version. Please update it in `package.json` and make noisy note of it in the readme with something like "BREAKING CHANGE: in adding TLRs we removed support for interfaces that only handle item-level requests."

Done! Thanks!

mkuklis (approved):
Nice work @Stanislau-Snapok!

Stanislau-Snapok:
> Nice work @Stanislau-Snapok!

This is the work of the whole team, not only mine, but thanks!

README.md
```diff
@@ -18,3 +18,7 @@ See project [UIREQ](https://issues.folio.org/browse/UIREQ)
 at the [FOLIO issue tracker](https://dev.folio.org/guidelines/issue-tracker).
 
 Other FOLIO Developer documentation is at [dev.folio.org](https://dev.folio.org/)
+
+## BREAKING CHANGES
+
+In adding title-level request we removed support for interfaces that only handle item-level requests.
```
zburke, line 24: Looking carefully at my last comment, I see I asked to update the readme with breaking change info. I meant the CHANGELOG; sorry!

zburke (approved):
This looks great!

README.md
Stanislau-Snapok: I can't fully understand how it should looks in `CHANGELOG`. Is it ok if I'll do something like this:

* BREAKING CHANGE: TLR - data migration for request list, view, create, edit, duplicate. Refs UIREQ-664.
* BREAKING CHANGE: TLR - data migration for dialogboxes and reorder list. Refs UIREQ-665.
* BREAKING CHANGE: TLR - data migration for hold shelf clearance report. Refs UIREQ-677.
* BREAKING CHANGE: TLR - data migration, move `holdingsRecordId`. Refs UIREQ-685.

Because now we don't have any `BREAKING CHANGE` section, and in my opinion it's not the best way if add it in the middle of changelog.
Could you please clarify?
Thanks!

Stanislau-Snapok: done, changed to:
BREAKING: TLR - data migration for request list, view, create, edit, duplicate. Refs UIREQ-664.
BREAKING: TLR - data migration for dialogboxes and reorder list. Refs UIREQ-665.
BREAKING: TLR - data migration for hold shelf clearance report. Refs UIREQ-677.
BREAKING: TLR - data migration, move `holdingsRecordId`. Refs UIREQ-685.

PR #806, opened 2021-12-02
UIREQ-614: update request results page

pavelspichonak:
## Purpose
Update request results page

## Refs
https://issues.folio.org/browse/UIREQ-614


Stanislau-Snapok (approved):

src/routes/utils.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  years = years
+    .filter((year) => isYear(year))
+    .map((year) => parseInt(year, 10));
+  years = orderBy(years, null, ['desc'])
+    .slice(0, DISPLAYED_YEARS_AMOUNT);
+
+  return years.join(YEAR_SEPARATOR);
+};
```
Dmitriy-Litvinenko, line 24: Could you please check possible write code something like this?

`export const getFormattedYears = (publications) => {
  const years = publications?.map(({ dateOfPublication }) => dateOfPublication);

  return years?.length
    ? ''
    : years
      .filter((year) => isYear(year))
      .map((year) => parseInt(year, 10))
      .sort(years, null, ['desc']) //
      .slice(0, DISPLAYED_YEARS_AMOUNT)
      .join(YEAR_SEPARATOR);
};`

Dmitriy-Litvinenko (approved):

src/routes/utils.js
pavelspichonak: Done.

mkuklis (approved):

PR #810, opened 2021-12-07
UIREQ-691: Closed TLR should not be able to be duplicated when TLR is disabled

Stanislau-Snapok:
## Purpose
Closed title-level request should not be able to be duplicated when TLR is disabled in settings.
Also if we have no any action for request,`Actions` button should not apear.

## Approach
We just need to check the level of `Closed` request and TLR feature state in settings.
If there are no buttons in action menu, the `Actions` button is not rendered by default.

## Refs
https://issues.folio.org/browse/UIREQ-691

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Create title level request checkbox. Refs UIREQ-655.
 * Update request results page. Refs UIREQ-614.
 * Update request details pane. Refs UIREQ-613.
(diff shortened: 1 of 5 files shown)


Dmitriy-Litvinenko (approved):

mkuklis (approved):
Nice and clean @Stanislau-Snapok!

pavelspichonak (approved):

PR #813, opened 2021-12-10
UIREQ-605 - Add preferred name to Requests UI

IvanKryzhanovskyi:
## Purpose
As a staff user
I want to see the patron's preferred first name in the name display in the Checkout and Request UI
so that I might address the patron properly.

## Approach
condition, included preferredFirstName has been added to the getFullName function

## Refs
https://issues.folio.org/browse/UIREQ-605


zburke (approved):

zburke:
Replaced by #814 since we can't run some CI processes on forks.

PR #827, opened 2021-12-17
UIREQ-630, UIREQ-644: Update queue page

Stanislau-Snapok:
## Purpose
Update queue page according to new requirements and mock ups

## Aproach
On new queue page we should have two separate accordions with different collections of fileds. We made separate component for each accordion. Inside we pass theire own map of fileds.
Additionaly we ecxtracted constans, which can be used for both accordions, in separate file.
Also we droped one `big-test` because we don't any more have the field, which was tested by it.

## Refs
https://issues.folio.org/browse/UIREQ-630
https://issues.folio.org/browse/UIREQ-644

## Screenshots
image


src/views/RequestQueueView.js
```diff
@@ -254,25 +208,49 @@ class RequestQueueView extends React.Component {
...
-      : '-';
+    const {
+      title,
+      publication,
+      contributors,
+    } = request?.instance || {};
+    const notYetFilledRequests = [];
+    const inProgressRequests = [];
+
+    requests.forEach(r => {
```
pavelspichonak, line 245: I believe it should be done once and should be stored in the state

pavelspichonak (approved):

src/views/RequestQueueView.js
Stanislau-Snapok: Yeah. I believe we should extract this logic from component. It will be done in scope of migration to new queue end-points. I am working on it right now.

Stanislau-Snapok:
@zburke, @mkuklis Could you review please?
Thanks!

mkuklis (approved):
Great work @Stanislau-Snapok as always!

src/routes/RequestsRoute.js
```diff
@@ -66,6 +66,7 @@ import { getFormattedYears } from './utils';
 
 const INITIAL_RESULT_COUNT = 30;
 const RESULT_COUNT_INCREMENT = 30;
+const DISPLAYED_YEARS_AMOUNT = 3;
```
mkuklis, line 69: @Stanislau-Snapok I see the same constant in `TitleInformation` perhaps it would be good to extract it to `constants` so it can be reused? Or perhaps the second argument in getFormattedYears can default to `DISPLAYED_YEARS_AMOUNT`.

Stanislau-Snapok: Were extracted into a global file with constants. Thanks!

PR #830, opened 2021-12-21
UIREQ-695: Migrate requests queue/reorder page on new end-points

Stanislau-Snapok:
## Purpose
Migrate requests queue/reorder page on new end-points

## Approach
Depends on is TLR swithced on or off we should pass instanceId or itemId in path and handle this inside component. Depends on what was passed, we should do fetch requests from correct endpoints, and send back reordered queue on correct end-point.

When we send back reorderd queue we should send full queue from first to last position, according this fact `notYetFilledRequests` and `inProgressRequests` collections was extracted one layer highe.

As we now have two separate accordions with requests but still one queue, reorder approach was a little bit changed.

Additionaly I removed `sort` of requests by position because new queue end-points return already sorted array.

Also fixed some small issues which can lead to errors or bugs.

## Refs
https://issues.folio.org/browse/UIREQ-695

### DIFF
--- a/src/PositionLink.js
+++ b/src/PositionLink.js
@@ -12,22 +12,23 @@ import {
 
 export default function PositionLink({ request }) {
   const queuePosition = get(request, 'position');
-  const barcode = get(request, 'item.barcode');
   const id = request[request.requestLevel === REQUEST_LEVEL_TYPES.ITEM ? 'itemId' : 'instanceId'];
   const openRequestsPath = `/requests?filters=${openRequestStatusFilters}&query=${id}&sort=${REQUEST_DATE}`;
 
-  return (request && barcode ?
-    <div>
-      <span>
-        {queuePosition}
-        &nbsp;
-        &nbsp;
-      </span>
-      <Link to={openRequestsPath}>
-        <FormattedMessage id="ui-requests.actions.viewRequestsInQueue" />
-      </Link>
-    </div> : '-'
-  );
+  return request
+    ? (
+      <div>
+        <span>
+          {queuePosition}
+          &nbsp;
+          &nbsp;
+        </span>
+        <Link to={openRequestsPath}>
+          <FormattedMessage id="ui-requests.actions.viewRequestsInQueue" />
+        </Link>
+      </div>
+    )
+    : '-';
 }
 
 PositionLink.propTypes = {
--- a/src/routes/RequestQueueRoute.js
+++ b/src/routes/RequestQueueRoute.js
@@ -3,13 +3,14 @@ import PropTypes from 'prop-types';
 import {
   get,
   keyBy,
-  sortBy,
 } from 'lodash';
 import ReactRouterPropTypes from 'react-router-prop-types';
 import { stripesConnect } from '@folio/stripes/core';
 
 import RequestQueueView from '../views/RequestQueueView';
 import urls from './urls';
+import { requestStatuses } from '../constants';
+import { getTlrSettings } from '../utils';
 
 class RequestQueueRoute extends React.Component {
   static getRequest(props) {
@@ -23,6 +24,14 @@ class RequestQueueRoute extends React.Component {
   }
 
   static manifest = {
+    configs: {
+      type: 'okapi',
+      records: 'configs',
+      path: 'configurations/entries',
+      params: {
+        query: '(module==SETTINGS and configName==TLR)',
+      },
+    },
     addressTypes: {
       type: 'okapi',
       path: 'addresstypes',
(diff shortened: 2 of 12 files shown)


src/routes/RequestQueueRoute.js
```diff
@@ -90,17 +106,76 @@ class RequestQueueRoute extends React.Component {
...
+        params,
+      },
+    } = this.props;
+
+    if ((prevConfigs.hasLoaded !== configs.hasLoaded && configs.hasLoaded) || prevConfigs.records[0]?.value !== configs.records[0]?.value) {
+      this.getRequests(configs, requests, params.id);
+    }
+  }
+
+  getRequests = (configs, requestsMutator, id) => {
```
pavelspichonak, line 158: looks like we do not need to pass requestsMutator and config, we can get them from props

src/routes/RequestQueueRoute.js
```diff
@@ -90,17 +106,76 @@ class RequestQueueRoute extends React.Component {
...
+    if (!configs.hasLoaded) {
+      return;
+    }
+
+    const { titleLevelRequestsFeatureEnabled } = getTlrSettings(configs.records[0]?.value);
+
+    this.setState({ titleLevelRequestsFeatureEnabled });
+    requestsMutator.reset();
+
+    if (titleLevelRequestsFeatureEnabled) {
```
pavelspichonak, line 168: const path = `circulation/requests/queue/${titleLevelRequestsFeatureEnabled ? 'instance' : 'item'}/${id}`;

requestsMutator.GET({path});

src/routes/RequestQueueRoute.js
```diff
@@ -151,11 +234,23 @@ class RequestQueueRoute extends React.Component {
     const { resources, location } = this.props;
     const request = this.getRequest();
     const requests = this.getRequestsWithDeliveryTypes();
+    const notYetFilledRequests = [];
+    const inProgressRequests = [];
+
+    requests
```
pavelspichonak, line 240: this logic should not be in render, let us think where to move it and save it to component state

src/views/components/ItemLink.js
```diff
@@ -14,11 +14,11 @@ const ItemLink = ({
     instanceId,
     holdingsRecordId,
     itemId,
-    item: { barcode },
+    item,
```
pavelspichonak, line 17: What did we remove it for?

src/routes/RequestQueueRoute.js
```diff
@@ -90,17 +106,76 @@ class RequestQueueRoute extends React.Component {
...
+      mutator: { requests },
+      resources: {
+        configs,
+      },
+      match: {
+        params,
+      },
+    } = this.props;
+
+    if ((prevConfigs.hasLoaded !== configs.hasLoaded && configs.hasLoaded) || prevConfigs.records[0]?.value !== configs.records[0]?.value) {
```
pavelspichonak, line 153: I would double check this `if` condition, looks too complicated. Maybe can be simplified.

src/PositionLink.js
```diff
@@ -12,22 +12,23 @@ import {
...
-        {queuePosition}
-        &nbsp;
-        &nbsp;
-      </span>
-      <Link to={openRequestsPath}>
-        <FormattedMessage id="ui-requests.actions.viewRequestsInQueue" />
-      </Link>
-    </div> : '-'
-  );
+  return request
```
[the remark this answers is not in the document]
Stanislau-Snapok: you are welcome

src/views/components/ItemLink.js
Stanislau-Snapok: When request is `Title` level, we have no item in it and it become `undefined`. There is an error when we will try to take `item.barcode`. But now `item.barcode` will be executed only when we guaranteed got `item`.

src/routes/RequestQueueRoute.js
Stanislau-Snapok: Good point! Thanks.

src/routes/RequestQueueRoute.js
```diff
@@ -90,17 +106,76 @@ class RequestQueueRoute extends React.Component {
       }),
       requests: PropTypes.shape({
         records: PropTypes.array,
+      }).isRequired,
+      configs: PropTypes.shape({
+        records: PropTypes.array.isRequired,
+        hasLoaded: PropTypes.bool.isRequired,
       }),
```
pavelspichonak, line 113: isRequired

src/views/RequestQueueView.js
```diff
@@ -62,10 +62,19 @@ class RequestQueueView extends React.Component {
...
+    } } = props;
+
+    const requestsPositionsForReorder = notYetFilledRequests.map(r => r.position);
 
     if (!state.reorder) {
-      return { requests };
+      return {
+        notYetFilledRequests,
+        inProgressRequests,
+        requestsPositionsForReorder,
```
pavelspichonak, line 76: Discussed => move this logic to constructor and componentDidUpdate

src/routes/RequestQueueRoute.js
Stanislau-Snapok: Done

src/routes/RequestQueueRoute.js
Stanislau-Snapok: Won't do as discussed

src/routes/RequestQueueRoute.js
Stanislau-Snapok: Done

src/routes/RequestQueueRoute.js
Stanislau-Snapok: Done

src/views/RequestQueueView.js
Stanislau-Snapok: Done

pavelspichonak (approved):

Stanislau-Snapok:
@mkuklis @zburke Could you please review?

mkuklis (approved):

Stanislau-Snapok:
We have some problems with our `bigtest`, but by reason that at curent moment we don't need this PR on snapshot and master branch, it will be merged without tests pass on pipeline.
There is the evidence of seccefull tests pass localy:
image

PR #838, opened 2021-12-27
UIREQ-706: Disable validation on reordering for `Page` requests for `TLR` feature

Stanislau-Snapok:
## Purpose
Disable validation on reordering for `Page` requests for `TLR` feature.

## Approach
We just need to check is our TLR feature enabled.
Also we can remove validation on not yet filled request, because now we have two accordions for requests with different statuses. At the second accordion, where we can reorder queue, will not be requests with status differ than `Open - Not yet filled`.

## Refs
https://issues.folio.org/browse/UIREQ-706

### DIFF
--- a/src/views/RequestQueueView.js
+++ b/src/views/RequestQueueView.js
@@ -58,6 +58,7 @@ class RequestQueueView extends React.Component {
       holding: PropTypes.object,
       request: PropTypes.object,
     }),
+    isTlrEnabled: PropTypes.bool.isRequired,
     onClose: PropTypes.func,
     onReorder: PropTypes.func,
     isLoading: PropTypes.bool,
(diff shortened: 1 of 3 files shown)


src/views/RequestQueueView.js
```diff
@@ -58,6 +58,7 @@ class RequestQueueView extends React.Component {
       holding: PropTypes.object,
       request: PropTypes.object,
     }),
+    isTlrEnabled: PropTypes.bool.isRequired,
```
pavelspichonak, line 61: should be optional or passed constant above should be like this const titleLevelRequestsFeatureEnabled = !!this.state?.titleLevelRequestsFeatureEnabled;

pavelspichonak (approved):

src/views/RequestQueueView.js
Stanislau-Snapok: Done, thanks!

PR #840, opened 2021-12-28
UIREQ-707: Add all required attributes for the `position` field in request info

Stanislau-Snapok:
## Purpose
Add all required attributes for the `position` field in request info

## Approach
End-points for geting requests was changed because our BE team says that using of `request-storage` is not correct way.
Removed multiple `settings` parsing, now all necessary fields stored in the `state`.

## Refs
https://issues.folio.org/browse/UIREQ-707

### DIFF
--- a/src/routes/RequestsRoute.js
+++ b/src/routes/RequestsRoute.js
@@ -42,6 +42,7 @@ import {
   requestTypesTranslations,
   REQUEST_LEVEL_TYPES,
   DEFAULT_DISPLAYED_YEARS_AMOUNT,
+  requestStatuses,
 } from '../constants';
 import {
   buildUrl,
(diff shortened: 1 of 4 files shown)


src/routes/RequestsRoute.js
```diff
@@ -109,6 +110,14 @@ class RequestsRoute extends React.Component {
   static contextType = CalloutContext;
 
   static manifest = {
+    configs: {
+      type: 'okapi',
+      records: 'configs',
+      path: 'configurations/entries',
```
pavelspichonak, line 116: do we really need to move it?

src/routes/RequestsRoute.js
```diff
@@ -595,6 +619,12 @@ class RequestsRoute extends React.Component {
         dynamicProperties.itemRequestCount = get(itemRequests, 'totalRecords', 0);
       }
 
+      if (titleLevelRequestsFeatureEnabled) {
```
pavelspichonak, line 622: const requests = titleLevelRequestsFeatureEnabled ? titleRequests : itemRequests;

dynamicProperties.numberOfNotYetFilledRequests = requests.requests.filter(request => request.status === requestStatuses.NOT_YET_FILLED).length;

pavelspichonak (approved):

src/routes/RequestsRoute.js
Stanislau-Snapok: Done, thanks!

src/routes/RequestsRoute.js
Stanislau-Snapok: You right, we don't need it. Done.

Stanislau-Snapok:
All tests are passed localy:
image

PR #841, opened 2021-12-29
UIREQ-700: Select item modal should display all items in the Instance…

pavelspichonak:
## Purpose
Select item modal should display all items in the Instance when user uncheck box

## Approach
Now we have **ItemsDialog** component which just render passed items.

This component is used in **move request dialog** and in **create request form**. Both these places get items from backend but **create request form** does not get all necessary items information(it is a bug) but **move request dialog** get all necessary items information.

So we decided:
1) remove all logic from **move request dialog** and **create request form** related to getting items
2) move this logic to **ItemsDialog** component(using correct logic from **move request dialog**)
3) add items sorting based on discussion with PO that items with status available must go first

## Refs
https://issues.folio.org/browse/UIREQ-700


Stanislau-Snapok (approved):

pavelspichonak:
@zburke @mkuklis could you please review?

src/ItemsDialog.js
```diff
@@ -58,16 +65,102 @@ export const formatter = {
...
+  const contentData = useMemo(() => {
+    let resultItems = items;
+
+    if (skippedItemId) {
+      resultItems = resultItems
+        .filter(item => skippedItemId !== item.id);
+    }
+
+    // items with status available must go first
+    return resultItems.sort((a, b) => {
```
mkuklis, line 152: @pavelspichonak I think you can shorter this with:

````js
return resultItems.sort(a => a?.status?.name === itemStatuses.AVAILABLE ? -1 : 1);
````

mkuklis (approved):
@pavelspichonak nice and clean refactoring! 

I just left a tiny little comment.

src/ItemsDialog.js
pavelspichonak: Done. Thanks

PR #849, opened 2022-01-11
STCOM-904 - Result List Revisions.

JohnC-80:
The problem: currently result lists in FOLIO suffer due to poor accessibility. Row-level block anchors is the direct cause.

The solution: render links to detail records within result listing cells and remove the block level anchors.

1. Use the exported `defaultMCLRowFormatter` from stripes-components rather than the built-in anchor formatter of `SearchAndSort`
2. Implement a local function to get the URL of each result item.
3. Render the link in a cell using the `<TextLink>` component from stripes-components.
4. use the `SearchAndSort` prop setting `resultRowClickHandlers={false}` to suppress `SearchAndSort`'s built-in click behavior.
5. supply your app's own `resultIsSelected` logic - in this PR, we hold the selected item in a state key that's set when the record link is clicked and cleared when the view record is dismissed.

image

### DIFF
--- a/src/routes/RequestsRoute.js
+++ b/src/routes/RequestsRoute.js
@@ -22,6 +22,8 @@ import {
   filters2cql,
   FormattedTime,
   MenuSection,
+  TextLink,
+  DefaultMCLRowFormatter,
 } from '@folio/stripes/components';
 import {
   deparseFilters,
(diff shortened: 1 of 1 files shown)


zburke (approved):
LGTM!

PR #854, opened 2022-01-26
UIREQ-658: Fulfillment Preference field not respecting user fulfillment preferences in edit mode

Stanislau-Snapok:
## Purpose
Fulfillment Preference field not respecting user fulfillment preferences in edit mode.

## Refs
https://issues.folio.org/browse/UIREQ-658


pavelspichonak (approved):

Dmitriy-Litvinenko (approved):

zburke (commented):
The `isEditForm()` check was clearly deliberately added, though I'm struggling to understand why. Still, I don't feel great about removing something if I don't understand why it was there in the first place. 

Does anybody have an inkling of why we might show different fulfillment options on the create and edit screens? Should we check with the PO on this?

zburke (approved):
Per PO comments on the ticket, the desired behavior is for create and update to have the same options, i.e. this PR does indeed implement the "correct" behavior.

mkuklis (approved):

PR #856, opened 2022-01-28
UIREQ-727: Fix `view requests in queue` link behaviour

Stanislau-Snapok:
## Purpose
The user should be redirected at the same page if click on the "view requests in queue" link from `request detail` page and from `edit` page.
Also if request has one of `closed` status we should display `dash` in queue field.

## Approach
On `edit` page we didn't pass `isTlrEnabled` in `PositionLink` component. Thats the reason why we always was redirected on `item` queue insead of `instance`.
I tried to pass `parrentResources` into the `RequestForm` component from `ViewRequest` to get TLR feature status directly in `RequestForm`, but this broke some logic.
So I passed just `isTlrEnabled` flag from `ViewRequest` thru `RequestForm` into `PositionLink` component.

## Refs
https://issues.folio.org/browse/UIREQ-727

### DIFF
--- a/src/RequestForm.js
+++ b/src/RequestForm.js
@@ -142,6 +142,7 @@ class RequestForm extends React.Component {
         GET: PropTypes.func.isRequired,
       }).isRequired,
     }).isRequired,
+    isTlrEnabledOnEditPage: PropTypes.bool,
   };
 
   static defaultProps = {
@@ -150,6 +151,7 @@ class RequestForm extends React.Component {
     optionLists: {},
     pristine: true,
     submitting: false,
+    isTlrEnabledOnEditPage: false,
   };
 
   constructor(props) {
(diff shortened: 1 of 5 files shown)


src/RequestForm.js
```diff
@@ -1100,6 +1101,7 @@ class RequestForm extends React.Component {
       },
       errorMessage,
       onCancel,
+      isTlrEnabled,
```
Dmitriy-Litvinenko, line 1104: Could you please check, can we add default value for current variable?

Dmitriy-Litvinenko (approved):

pavelspichonak (approved):

src/RequestForm.js
Stanislau-Snapok: done

mkuklis (approved):

PR #857, opened 2022-01-31
UIREQ-730: Add link for `Requests on item` field if level of request is `Title`

Stanislau-Snapok:
## Purpose
Add link for `Requests on item` field if level of request is `Title`.

## Approach
We should not rely on request `level` because `title` requests in some moment can have item in their record.
According to discussing with PO, we should display `0` if have no requests on item (such case can be for closed requests), and this value should not be a `link`.

## Refs
https://issues.folio.org/browse/UIREQ-730

### DIFF
--- a/src/ItemDetail.js
+++ b/src/ItemDetail.js
@@ -17,6 +17,8 @@ import {
 
 import { openRequestStatusFilters } from './utils';
 
+const DEFAULT_COUNT_VALUE = 0;
+
 const ItemDetail = ({
   currentInstanceId,
   request,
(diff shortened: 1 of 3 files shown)


src/ItemDetail.js
```diff
@@ -34,18 +34,20 @@ const ItemDetail = ({
   const holdingsRecordId = request?.holdingsRecordId || item.holdingsRecordId;
   const title = request?.instance.title || item.title || <NoValue />;
   const contributor = request?.instance.contributorNames?.[0]?.name || item.contributorNames?.[0]?.name || <NoValue />;
-  const count = request?.itemRequestCount || requestCount;
+  const count = request?.itemRequestCount || requestCount || 0;
```
Dmitriy-Litvinenko, line 37: Could you please move 0 to const?

src/routes/RequestsRoute.js
```diff
@@ -625,14 +629,14 @@ class RequestsRoute extends React.Component {
       const dynamicProperties = {};
       const requestsForFilter = titleLevelRequestsFeatureEnabled ? titleRequests : itemRequests;
 
-      dynamicProperties.numberOfNotYetFilledRequests = requestsForFilter.requests.filter(request => request.status === requestStatuses.NOT_YET_FILLED).length;
+      dynamicProperties.numberOfNotYetFilledRequests = requestsForFilter.requests.filter(r => r.status === requestStatuses.NOT_YET_FILLED).length;
```
Dmitriy-Litvinenko, line 632: Could you please rename r?

Dmitriy-Litvinenko (approved):

pavelspichonak (approved):

src/ItemDetail.js
Stanislau-Snapok: done

src/routes/RequestsRoute.js
Stanislau-Snapok: done

mkuklis (approved):

PR #864, opened 2022-02-11
UIREQ-739 candidate...

JohnC-80:
See https://issues.folio.org/browse/UIREQ-739 about pane size truncation.

Panes use the pane's id to track their widths in the resizing logic... if an id is not provided, a random number id is generated. This can give the resize logic some issues.

This PR assigns an id to the requests detail `<Pane>` AND the spinner-rendering `<Pane>` which might at least lend stability if it doesn't fix the issue.

Could someone please try this out to see if it resolves? thanks.
The centralized fix is on a branch of stripes-components: https://github.com/folio-org/stripes-components/pull/1721

### DIFF
--- a/src/ItemDetail.js
+++ b/src/ItemDetail.js
@@ -37,7 +37,7 @@ const ItemDetail = ({
   const title = request?.instance.title || item.title || <NoValue />;
   const contributor = request?.instance.contributorNames?.[0]?.name || item.contributorNames?.[0]?.name || <NoValue />;
   const count = request?.itemRequestCount || requestCount || DEFAULT_COUNT_VALUE;
-  const status = item.status.name || item.status || <NoValue />;
+  const status = item.status?.name || item.status || <NoValue />;
   const effectiveLocationName = item.effectiveLocation?.name || item.location?.name || <NoValue />;
   const dueDate = loan?.dueDate ? <FormattedDate value={loan.dueDate} /> : <NoValue />;
 
--- a/src/ViewRequest.js
+++ b/src/ViewRequest.js
@@ -569,6 +569,7 @@ class ViewRequest extends React.Component {
         dismissible
         {... (showActionMenu ? { actionMenu } : {})}
         onClose={this.props.onClose}
+        id="request-detail-pane"
       >
         <TitleManager record={get(request, ['instance', 'title'])} />
         <AccordionSet accordionStatus={accordions} onToggle={this.onToggleSection}>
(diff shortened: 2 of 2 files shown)


pavelspichonak:
Hi @JohnC-80. I've tested fix from [your pull request](https://github.com/folio-org/stripes-components/pull/1721) and could not reproduce [this bug](https://issues.folio.org/browse/UIREQ-739) anymore.

And looks like bug is not related to **id** property. So I believe we can close this pull request and merge [this pull request](https://github.com/folio-org/stripes-components/pull/1721)

Sorry for the long wait and thank you!

pavelspichonak:
@JohnC-80 oh now I understand why we need to pass ids, local storage is used to cache pane size. Thanks for clarification. I will discuss it with our team tomorrow, I believe we will create some story because we have a lot of places in different modules where pane is used without id.

JohnC-80:
@pavelspichonak After talking it over with some `stripes` developers, I decided to work up a centralized solution that simply wouldn't cache the pane widths if an id was not provided. This will protect apps/panesets from inadvertently growing local storage any in case they haven't provided an id. to their panes. https://github.com/folio-org/stripes-components/pull/1725

pavelspichonak:
Hi @JohnC-80. We have created tech debt tasks to add ids to **Pane** component in all places where it is used in modules we are responsible for. I believe we can close this pull request, in future we will create pull request in which we add **ids** in all places in **ui-requests** module.

I would also suggest if some common component uses **Pane** component we should not assign **id** property to **Pane** in such common components because common components can be used multiple times in different places in application and all of them will have the same **ids** and will affect each other.

For example we have **SearchAndSort** smart component which uses **Pane** component. **SearchAndSort** component is used multiple times in different modules. Maybe we should pass **ids** to **SearchAndSort** component and **SearchAndSort** will pass these **ids** to **Pane** component?

Also I'd suggest to use **ids** with namespaces something like this `id="ui-requests.requestsList.filtersColumn"` to avoid duplicated **ids**.

Thank you one more time!

pavelspichonak:
Hi @JohnC-80. Thank you for your effort. We have created pull requests for each module to fix issue related to **Pane** component missing **ids**
[https://issues.folio.org/browse/UICHKIN-329](https://issues.folio.org/browse/UICHKIN-329)
[https://issues.folio.org/browse/UICIRC-756](https://issues.folio.org/browse/UICIRC-756)
[https://issues.folio.org/browse/UICHKOUT-768](https://issues.folio.org/browse/UICHKOUT-768)
[https://issues.folio.org/browse/UIREQ-742](https://issues.folio.org/browse/UIREQ-742)

**Ids** will be added when we start working on these tasks. I close this pull request.

PR #866, opened 2022-02-16
UIREQ-675: Add title look-up to new request form

Stanislau-Snapok:
## Purpose
We need to add find-instance plugin on create request page for `title` level requests.

## Refs
https://issues.folio.org/browse/UIREQ-675

### DIFF
--- a/package.json
+++ b/package.json
@@ -210,6 +210,7 @@
     "react-router-dom": "^5.2.0"
   },
   "optionalDependencies": {
-    "@folio/plugin-find-user": "^6.0.0"
+    "@folio/plugin-find-user": "^6.0.0",
+    "@folio/plugin-find-instance": "*"
   }
 }
--- a/package.json
+++ b/package.json
@@ -211,6 +211,6 @@
   },
   "optionalDependencies": {
     "@folio/plugin-find-user": "^6.0.0",
-    "@folio/plugin-find-instance": "*"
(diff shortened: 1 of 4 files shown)


Dmitriy-Litvinenko (approved):

package.json
```diff
@@ -210,6 +210,7 @@
     "react-router-dom": "^5.2.0"
   },
   "optionalDependencies": {
-    "@folio/plugin-find-user": "^6.0.0"
+    "@folio/plugin-find-user": "^6.0.0",
+    "@folio/plugin-find-instance": "*"
```
Stanislau-Snapok, line 214: Hello @zburke. Could you please clarify, should we change here plugin version after our changes for it will be released? Thanks!

zburke: IMO `*` is never an acceptable dependency. I would expect a `^6` dep would protect you here, and allow your application to receive the newest-available version, even if it's not released yet, since I anticipate only minor/patch releases for that plugin in Lotus.

zburke (changes requested):
Please use a more conservative dependency constraint in your `package.json`. Otherwise, this looks good.

package.json
Stanislau-Snapok: Done

zburke (approved):

mkuklis (approved):

PR #880, opened 2022-03-10
UIREQ-747: Add success toast to duplicated Requests

pavelspichonak:
## Purpose
Add success toast to duplicated Requests

## Refs
https://issues.folio.org/browse/UIREQ-747


Stanislau-Snapok (approved):

mkuklis (approved):

zburke (changes requested):
lint that `:` 🙄 and then this looks great!

Stanislau-Snapok:
> lint that `:` 🙄 and then this looks great!

Good point! Pavel is on vacation now, so I'll take care about this. Thanks!

PR #884, opened 2022-03-23
UIREQ-753 fix notification issue with first name

artem-blazhko:
## Purpose

Fix issue with success toast when user did not save first name

## Refs

[UIREQ-753](https://issues.folio.org/browse/UIREQ-753)
## Screenshots

BEFORE

image

AFTER

Success toast with first name and last name

image

Success toast with last name only 

image

Success toast with full name (including middle name)

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Add title look-up to new request form. Refs UIREQ-675.
 * Add success toast to Requests. Refs UIREQ-722.
 * Add success toast to duplicated Requests. Refs UIREQ-747.
+* Fix defect with first name of success toast to Requests. Refs UIREQ-753. 
 
 ## [7.0.0](https://github.com/folio-org/ui-requests/tree/v7.0.0) (2022-02-25)
 [Full Changelog](https://github.com/folio-org/ui-requests/compare/v6.0.2...v7.0.0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 * Move page requests to first accordion of unified queue when reordering. Refs UIREQ-728.
 * Add title look-up to new request form. Refs UIREQ-675.
 * Add success toast to Requests. Refs UIREQ-722.
-* Add success toast to duplicated Requests. Refs UIREQ-747. 
+* Add success toast to duplicated Requests. Refs UIREQ-747.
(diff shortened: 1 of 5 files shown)


Stanislau-Snapok (approved):

zburke (changes requested):
Please use `getFullName` from [@folio/stripes-util](https://github.com/folio-org/stripes-util/blob/3cea0ec7af6a5472ecd1d64d8c31cb15f8d420d9/lib/getFullName.js).

artem-blazhko:
Hello @zburke. 
I can not use _getFullName_ method from stripes-util library.  
In our case, we have a bit different requirements that can not be achieved by using library method.
We have case when a user has lastName and middleName but middleName should not be shown.

In general, there is a list of possible cases: 

last name (no first or middle name present)
last name, first name (no middle name present)
last name, first name middle name

artem-blazhko:
> Hello @zburke. I can not use _getFullName_ method from stripes-util library. In our case, we have a bit different requirements that can not be achieved by using library method. We have case when a user has lastName and middleName but middleName should not be shown.
> 
> In general, there is a list of possible cases:
> 
> last name (no first or middle name present) last name, first name (no middle name present) last name, first name middle name

@zburke could you take a look?

zburke (approved):

zburke:
The function itself looks fine of course. To me, the real question is, “Are we confident FOLIO wants to use different algorithms for displaying ‘Last, First Middle’ [all](https://github.com/folio-org/ui-inventory/blob/474ab4e7a2ade6419ad80c10d97b77dd58ca13d6/src/views/ItemView.js#L430) (ui-users) [over](https://github.com/folio-org/ui-users/blob/9d9e32cf7654ce6f53b60b7d3c2471bd8112837d/src/components/util/util.js#L11-L32) (ui-inventory) [the](https://github.com/folio-org/stripes-util/blob/master/lib/getFullName.js) (stripes-util) [place](https://github.com/folio-org/stripes-smart-components/blob/7e97f68894253f7e4abb225295b2ae75b70d38bb/lib/UserName/UserName.js) (stripes-smart-components)?” But, that’s not really your problem and this is fine for now. 

Filed [STRIPES-789](https://issues.folio.org/browse/STRIPES-789) if you want to follow that discussion.

PR #885, opened 2022-03-24
UIREQ-754: fix request details `# items` to show correct number

Stanislau-Snapok:
## Purpose
There was mismatching between number of requests on `request detail` page and reorderable accordion on the `reorder queue` page.

## Approach
Again we can see here the problem with query limits witch is `10` by default. 

## Refs
https://issues.folio.org/browse/UIREQ-754


artem-blazhko (approved):

zburke (approved):
Good catch! I think the default `limit=10` parameter in RMB may be the sneakiest/most-problematic decision I have even encountered.

PR #888, opened 2022-03-24
UIREQ-751: fix the issue when pickup location was not preserved on request duplication

Stanislau-Snapok:
## Purpose
When duplicating a request, the pickup location is not preserved

## Approach
According to code in the `duplicateRequest` function, we omit `requesterId` field, and than in `RequestForm.js` we trying to compare it and it's equal to `undefined`:
`
if (requesterId !== selectedUser?.id || this.props?.query?.mode !== createModes.DUPLICATE) {
        change('pickupServicePointId', defaultServicePointId);
}
`

## Refs
https://issues.folio.org/browse/UIREQ-751

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * Add success toast to Requests. Refs UIREQ-722.
 * Add success toast to duplicated Requests. Refs UIREQ-747.
 * Change queue position message from "items" to "requests". Refs UIREQ-755.
+* When duplicating a request, the pickup location is not preserved. Refs UIREQ-751.
 
 ## [7.0.0](https://github.com/folio-org/ui-requests/tree/v7.0.0) (2022-02-25)
 [Full Changelog](https://github.com/folio-org/ui-requests/compare/v6.0.2...v7.0.0)
--- a/src/utils.js
+++ b/src/utils.js
@@ -120,7 +120,6 @@ export function duplicateRequest(request) {
     'proxyUserId',
     'requestCount',
     'requester',
-    'requesterId',
     'status',
   ]);
 }
--- a/src/utils.test.js
+++ b/src/utils.test.js
@@ -108,7 +108,6 @@ describe('duplicateRequest', () => {
       'proxyUserId',
       'requestCount',
       'requester',
-      'requesterId',
       'status',
     ];
 


zburke (changes requested):
`requesterId` was very deliberately added to the omit-list in #762 to resolve a problem with duplicating requests made by proxies. Please follow up with @Baroquem. We likely need to find an alternate way to preserve the pickup-location without also preserving stale proxy information.

Stanislau-Snapok:
> `requesterId` was very deliberately added to the omit-list in #762 to resolve a problem with duplicating requests made by proxies. Please follow up with @Baroquem. We likely need to find an alternate way to preserve the pickup-location without also preserving stale proxy information.

@zburke I checked this behaviour on the snapshot and localy with this branch. What I did:
1. Set Morty Panic as a proxy for Rick Psych
2. Create a request where Morty act as self.
3. Try to duplicate this request and now act as proxy for Rick.
4. Change user to Sheldon, Bazinga (he have no any proxies or sponsors)
5. Select pickup service point.
6. Try to save request and get the error:
image

I checked this behaviour by a lot of different ways, but got error only in this case. The same behaviour i see on my local branch with that exception that pickup service point is preserved when we are trying to duplicate request.
For me it looks like my changes doesn't break anything.
Am I right or I miss something?
Thanks!

Stanislau-Snapok:
Hello @Baroquem. Could you please take a look?

Baroquem:
Hi, @Stanislau-Snapok ... I apologize for not reviewing this PR sooner. Unfortunately, I think we have a couple of issues tangled up here. I am able to confirm what you demonstrated above, that the `proxyUserId is not valid` error is present on snapshot. However, @zburke is correct in saying that the work I did with `requesterId` was done to fix that error. This suggests that there has either been a regression at some point, causing the error to appear again, or else that you've uncovered a slightly different variation on the original error. It's difficult to reproduce the original error report (https://issues.folio.org/browse/UIREQ-626) because it refers to specific users and items in an old bugfest environment, so I'm not sure which case we're dealing with here.

It's not your fault, since the error was already present in snapshot, but it is possible that the work you've done here would interfere with the earlier work. After discussing this with Zak and others, I think we need to get some guidance from the POs about how this duplication behavior should work before trying to merge the current PR.

@bltravis , would you mind taking a look at this proxy error again in snapshot and helping us work out what's going on? Specifically, can you recall whether the steps Stanislau followed in his comment a few days ago are the same that you used in the description of UIREQ-626 (indicating a regression bug) or a new aspect of duplication with proxy that we may not have considered last time?

bltravis:
Sorry, folks. I hadn’t seen the request for my input. I’ll take a look at this ASAP, but we can definitely defer this if we need to focus elsewhere.

PR #890, opened 2022-03-29
UIREQ-757: changes requests filter logic to  shows correct number in `# requests`

Stanislau-Snapok:
## Purpose
Ensure Request details # (# requests) shows correct data.

## Approach
After changes that was made in [UIREQ-728](https://issues.folio.org/browse/UIREQ-728) all `Page` requests situated in non-reorderable accordion. As discussed with PO, our `# reuqests` should represent number of reorderable requests.
I added coresponding filter and also renamed `numberOfNotYetFilledRequests` to `numberOfReorderableRequests` because it's no more match with what exact this number mean.

## Refs
https://issues.folio.org/browse/UIREQ-757

### DIFF
--- a/src/routes/utils.js
+++ b/src/routes/utils.js
@@ -32,9 +32,5 @@ export const getFormattedContributors = (contributors) => (
 );
 
 export const isReorderableRequest = request => {
-  if (request.status === requestStatuses.NOT_YET_FILLED && request.requestType !== requestTypesMap.PAGE) {
-    return true;
-  }
-
-  return false;
+  return request.status === requestStatuses.NOT_YET_FILLED && request.requestType !== requestTypesMap.PAGE;
 };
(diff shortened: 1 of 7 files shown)


artem-blazhko (approved):

src/routes/utils.js
```diff
@@ -25,3 +30,11 @@ export const getFormattedPublishers = (publications) => (
 export const getFormattedContributors = (contributors) => (
   contributors?.find(({ name }) => !!name)?.name ?? ''
 );
+
+export const isReorderableRequest = request => {
+  if (request.status === requestStatuses.NOT_YET_FILLED && request.requestType !== requestTypesMap.PAGE) {
```
pavelspichonak, line 35: looks like just `return request.status === requestStatuses.NOT_YET_FILLED && request.requestType !== requestTypesMap.PAGE`

pavelspichonak (approved):

src/routes/utils.js
Stanislau-Snapok: done

PR #910, opened 2022-05-13
UIREQ-775: Create/Edit Request: required is not read

Dmitriy-Litvinenko:


Dmitriy-Litvinenko:
After discussion with other developer we decide that approach https://github.com/folio-org/ui-requests/pull/914 better

PR #914, opened 2022-05-17
UIREQ-775: Create/Edit Request: required is not read

Dmitriy-Litvinenko:
## Purpose
Create/Edit Request: required is not read

## Approach
For resolve this problem we add minimal changes in UI that was approved by PO
UIREQ-775

## Refs
https://issues.folio.org/browse/UIREQ-775

### DIFF
--- a/src/RequestForm.js
+++ b/src/RequestForm.js
@@ -1130,10 +1130,6 @@ class RequestForm extends React.Component {
 
     const isEditForm = this.isEditForm();
 
-    const labelAsterisk = isEditForm
-      ? ''
-      : ' *';
-
     const disableRecordCreation = true;
 
     let deliveryLocations;
@@ -1200,6 +1196,7 @@ class RequestForm extends React.Component {
         >
           <form
             id="form-requests"
+            noValidate
             className={css.requestForm}
             onSubmit={handleSubmit(this.onSave)}
             data-test-requests-form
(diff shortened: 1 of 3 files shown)


artem-blazhko (approved):

pavelspichonak (approved):

artem-blazhko (approved):

mkuklis (approved):

src/RequestForm.js
```diff
@@ -1374,26 +1366,30 @@ class RequestForm extends React.Component {
                                         name="item.barcode"
                                         placeholder={placeholder}
                                         aria-label={<FormattedMessage id="ui-requests.item.barcode" />}
+                                        label={<FormattedMessage id="ui-requests.item.barcode" />}
```
zburke, line 1369: Please omit `aria-label` now that this field properly has `label`.

src/RequestForm.js
```diff
@@ -1296,25 +1288,29 @@ class RequestForm extends React.Component {
                                           name="instance.hrid"
                                           placeholder={placeholder}
                                           aria-label={<FormattedMessage id="ui-requests.instance.value" />}
+                                          label={<FormattedMessage id="ui-requests.instance.value" />}
```
zburke, line 1291: Please omit `aria-label` now that this field properly has a `label`.

src/RequestForm.js
```diff
@@ -1586,12 +1578,14 @@ class RequestForm extends React.Component {
                                     name="requester.barcode"
                                     placeholder={placeholder}
                                     aria-label={<FormattedMessage id="ui-requests.requester.barcode" />}
+                                    label={<FormattedMessage id="ui-requests.requester.barcode" />}
```
zburke, line 1581: Please omit `aria-label` now that this field properly has a `label`.

zburke (changes requested):
Fields should have either `label` or `aria-label`, but not both.

src/RequestForm.js
Dmitriy-Litvinenko: Fixed

src/RequestForm.js
Dmitriy-Litvinenko: Fixed

PR #928, opened 2022-06-02
UIREQ-783: disable request detail action menu when hardcoded ID is present

Stanislau-Snapok:
## Purpose
Disable request detail action menu when hardcoded UID is present.

## Approach
On real envs when we switching on our title-level request feature, on back-end side migration script process all requests to create new queue. There would be a case, when migration script can't find real `instanceId` or `holdingsRecordId` or both of them. To not break anything within script execution in such cases back-end replaces problematic ids with `00000000-0000-4000-8000-000000000000` which is equal to our `INVALID_REQUEST_HARDCODED_ID`. On front-end side we should catch those invalid requests and disable all actions for them except `Close request`.

## Refs
https://issues.folio.org/browse/UIREQ-783


zburke (approved):
LGTM

PR #930, opened 2022-06-03
UIREQ-786: Fix big tests errors related to interactors small default …

pavelspichonak:
## Purpose
Fix big tests errors related to interactors small default timeout

## Refs
https://issues.folio.org/browse/UIREQ-786

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 * Disable request detail action menu when hardcoded UID is present. Refs UIREQ-783.
 * Disable item and instance links in request detail when hardcoded UID is present. Refs UIREQ-784.
 * Fix "Timeout of 2000ms exceeded. For async tests and hooks, ensure “done()” is called" big tests errors. Refs UIREQ-785.
+* Fix big tests errors related to interactors small default timeout. Refs UIREQ-786.
  
 ## [7.0.2](https://github.com/folio-org/ui-requests/tree/v7.0.2) (2022-04-04)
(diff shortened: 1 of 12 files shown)


Dmitriy-Litvinenko (approved):

Stanislau-Snapok (approved):

zburke (approved):
I have certainly been bitten by the 2000 millisecond timeout in bigtest, but changing all these tests is definitely a brute-force approach that also feels irksome. If it's a quick fix that buys you time to get things ported to Jest/RTL, sure, go for it. But if these tests are going to stick around for a while, it may be worth exploring why they are so slow.

pavelspichonak:
> I have certainly been bitten by the 2000 millisecond timeout in bigtest, but changing all these tests is definitely a brute-force approach that also feels irksome. If it's a quick fix that buys you time to get things ported to Jest/RTL, sure, go for it. But if these tests are going to stick around for a while, it may be worth exploring why they are so slow.

@zburke very often when we create pull requests some big tests randomly fail. And sometimes for example in the morning you can run tests and they are successful and then you run them in the evening and they start failing. So looks like problem is not related to source code. Locally I can see failing tests when in the network tab in dev tools I use 'fast 3G' internet speed. But as far as I understood in big tests real backend is not used so looks like issue is not related to internet speed either. So maybe it is related to high workload of machine where big tests run.

I agree that it is not good approach to fix them like this because we will not be able to find slow places in our application in future. But I hope soon we will have jest/RTL tests instead of big tests. I'll discuss with our team this topic one more time. Thanks!

mkuklis (approved):

PR #931, opened 2022-06-07
UIREQ-787: filter the items list on `move request` action from items with non-requestable statuses

Stanislau-Snapok:
## Purpose
Filter the items list on "move request" action from items with non-requestable statuses.

## Approach
When we have `skippedItemId` it means that we in `move request` process, so we should filter our list from items with non-requestable statuses. 

## Refs
https://issues.folio.org/browse/UIREQ-787


Dmitriy-Litvinenko (approved):

artem-blazhko (approved):

pavelspichonak (approved):

Stanislau-Snapok:
Hello @zburke, @mkuklis! Could you please review? Thanks!

PR #936, opened 2022-06-24
UIREQ-790: fix the number of title-level requests

Stanislau-Snapok:
## Note
Eslint fixed paddings, so it looks like a lot of changes, but in real only two strings was changed.
Without whitespaces: https://github.com/folio-org/ui-requests/pull/936/files?w=1

## Purpose
Fix the number of title-level requests on request details and within creation of new title-level request.

## Approach
The easiest way to get only title-level request is not working as expected. Relying on this number we calculate the number of reorderable requests. So had to just filter requests on title from requests with the same title but with `Item requestLevel`.

## Refs
https://issues.folio.org/browse/UIREQ-790


artem-blazhko (approved):

Dmitriy-Litvinenko (approved):

pavelspichonak (approved):

zburke (approved):
LGTM

PR #945, opened 2022-07-19
UIREQ-792: Components are incorrectly imported directly from stripes-* packages

Dmitriy-Litvinenko:
## Purpose
Components are incorrectly imported directly from stripes-* packages

## Refs
https://issues.folio.org/browse/UIREQ-792

## Screenshots
We have imports "@folio/stripes-" but all of them associated with tests
UIREQ-792


artem-blazhko (approved):

zburke (approved):
Thank you!

As a rule, PR-titles and CHANGELOG entries should describe the change you made rather than what problem you solved, e.g. "Correctly import components from @folio/stripes/* packages". A good way to write this is to complete the sentence, "After this PR, the code will ____." You don't need to change anything here, but it's sth to consider for next time.

Dmitriy-Litvinenko:
Fixed

PR #952, opened 2022-08-02
Use 7.1.0 version of stripes package

artem-blazhko:


artem-blazhko:
Closed after testing different approaches that should fix the build of b7.0 branch

PR #957, opened 2022-08-16
UIREQ-818: Render search results as normal and show dash for empty st…

Dmitriy-Litvinenko:
## Purpose
Render search results as normal and show dash for empty status

## Refs
https://issues.folio.org/browse/UIREQ-818

## Screenshots
On screen shot we can see difference between current behavior and future behavior after fix
UIREQ-818
UIREQ-818_2


artem-blazhko (approved):

zburke (approved):
LGTM. Thank you!

PR #975, opened 2022-11-04
UIREQ-824: TLR: request type reverts to page when placing request

artem-blazhko:
## Purpose
After selecting the request type onBlur event was executed. After that validation for some fields was called. It made reset of request type.

## Approach
I have deleted `validateOnBlur=true` property of Form. Instead of that I have implemented field level validation.
Also, in the scope of this PR I have moved some methods of RequestForm component to utils to make testing of the component easier in the future.

## Refs
[UIREQ-824](https://issues.folio.org/browse/UIREQ-824)

### DIFF
--- a/src/RequestForm.js
+++ b/src/RequestForm.js
@@ -1009,7 +1009,6 @@ class RequestForm extends React.Component {
   handleChangeItemBarcode = (event) => {
     const {
       form,
-      onSetSelectedItem,
     } = this.props;
     const {
       isItemBarcodeClicked,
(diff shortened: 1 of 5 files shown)


Dmitriy-Litvinenko (approved):

zburke (approved):
Wow, this is a big refactor! I completely agree that lifting all those utility functions out of the component to simplify testing is a great move. It's too bad SonarCloud doesn't give you credit for that; I wonder if it's still configured to calculate coverage from BigTest instead.

mkuklis (approved):

mkuklis:
Nice cleanup @artem-blazhko!

PR #976, opened 2022-11-09
UIREQ-837: Cannot submit a request with fulfillment preference = Delivery

artem-blazhko:
## Purpose
User is not able to select Fulfillment preference correctly due to incorrect migration from Redux Form to React Final Form. 
Also, the same issue with Delivery address selector.

## Refs
[UIREQ-837](https://issues.folio.org/browse/UIREQ-837)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Requests: Implement App context menu and keyboard shortcuts modal. refs UIREQ-817.
 * Make title level request box selected when duplicating title level request. Refs UIREQ-827.
 * Prevent request to get instance if there is no `instanceId`. Refs UIREQ-826.
+* Make Fulfillment preference and Delivery address selectable. Refs UIREQ-837. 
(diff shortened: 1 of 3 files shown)


Dmitriy-Litvinenko (approved):

zburke (commented):
@artem-blazhko, This PR fixes a relatively severe bug. Thank you! 

I'm impressed (and happy) that you were able to understand that the problem here was with the redux-form/final-form migration. I also see that test coverage for RequestForm.js (by RequestForm.test.js) is barely 30%. Do you think we could have caught it with better testing ahead of time? Can we add tests now to confirm that these settings are being correctly persisted?

mkuklis (approved):

artem-blazhko:
Thank you @zburke.
I added a few simple tests.
Also, I want to add that we have plans to split `RequestForm` component into a few smaller components. And of course, we will cover everything by tests.

PR #978, opened 2022-11-16
UIREQ-840: Cannot save item request from On Order item, because need a barcode

artem-blazhko:
## Purpose
When the user tries to create a request with `itemId` but **without** `item.barcode` we have a validation error for `item barcode` field which says that the field is empty. But we should not have that error. 
The issue appeared after migration from Redux Form to React Final Form.

## Approach
I have disabled validation of `item barcode` field if it is empty and `itemId` is presented (the same functionality as on Morning Glory).

## Refs
[UIREQ-840](https://issues.folio.org/browse/UIREQ-840)

### DIFF
--- a/src/RequestForm.js
+++ b/src/RequestForm.js
@@ -670,7 +670,7 @@ class RequestForm extends React.Component {
     });
   }
 
-  findItem(key, value, isValidation = false, disableValidation = true) {
+  findItem(key, value, isValidation = false, isBarcodeRequired = true) {
     const {
       findResource,
       form,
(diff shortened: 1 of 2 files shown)


Dmitriy-Litvinenko (approved):

src/RequestForm.js
```diff
@@ -208,6 +208,7 @@ class RequestForm extends React.Component {
       shouldValidateInstanceId: false,
       isInstanceIdBlur: false,
       validatedInstanceId: null,
+      isItemIdRequest: Boolean(this.props.query.itemId),
```
mkuklis, line 211: @artem-blazhko is `query` prop always present on `RequestForm` or would it be safer to do `this.props.query?.itemId` here?

mkuklis (approved):

src/RequestForm.js
artem-blazhko: Thank you @mkuklis.
I am not definitely sure that `query` is always present. But anyway I added conditional property.

PR #981, opened 2022-11-21
Fix issue with big tests during deploy of release

artem-blazhko:
## Purpose
During the **release** deploy (buildNPM Release) we have tests related errors:
https://github.com/folio-org/ui-requests/actions/runs/3512545663/jobs/5885459044

It happens because the module can not be found.
As a result, deployment fails.
The root cause of the issue is not clear. 
However, _buildNPM Snapshot_ deploy passed successfully.
    

## Approach 
I have created local files and removed imports.

## Refs
[UIREQ-847](https://issues.folio.org/browse/UIREQ-847)


Dmitriy-Litvinenko (approved):

zburke (commented):
IIUC, tests suddenly stopped working. This suggests a change either in stripes-webpack or stripes-smart-components had unintended side-effects. Have you looked for a guilty PR in either of those repositories? Let's look in those repos first and IFF we can't find a cause there then merge this. 

Generally, I am not in favor of doing anything to prolong BigTest's life. I understand you're not going to drop everything, skip this PR, and write a million Jest/RTL tests for ui-requests, but this should be a loud reminder that BigTest is deprecated and the work to migrate away should have been scheduled literally years ago.

artem-blazhko:
PR is not relevant and will be closed.

zburke:
The problem with the release was a mis-configured `latest` dist-tag on stripes-core in the npm-folio repo causing multiple copies of it to be pulled into the build. Once we fixed the config glitch, the build succeeded and it became clear that this change was unnecessary.

From the debugging discussion:
> The problem here was the misplaced latest dist-tag on stripes 7.2.4 even though stripes 7.3.1 was also available. This caused yarn to pull in stripes 7.2.4 (along with its Lotus-era ~stripes deps, including stripes-core 8.2.2) instead of stripes 7.3.1. Because you also have a dev dep for stripes-core ^8.0.0, this caused yarn to pull in stripes-core 8.3.0. With multiple copies of stripes-core in the bundle, tests would not run correctly.

PR #982, opened 2022-11-21
UIREQ-834: TLR: Inconsistent title info loading when request created from Instance record

artem-blazhko:
## Purpose
When the user tries to create a request from Inventory item information is not loaded the first time. But when the user refreshes the page everything works as expected.

## Refs
[UIREQ-834](https://issues.folio.org/browse/UIREQ-834)

### DIFF
--- a/src/routes/RequestsRoute.js
+++ b/src/routes/RequestsRoute.js
@@ -5,7 +5,10 @@ import {
 } from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
-import { stringify } from 'query-string';
+import {
+  stringify,
+  parse,
+} from 'query-string';
 import moment from 'moment-timezone';
 import {
   FormattedMessage,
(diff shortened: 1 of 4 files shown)


src/routes/RequestsRoute.js
```diff
@@ -402,7 +405,9 @@ class RequestsRoute extends React.Component {
       locale: PropTypes.string.isRequired,
     }).isRequired,
     history: PropTypes.object,
-    location: PropTypes.object.isRequired,
+    location: PropTypes.shape({
+      search: PropTypes.string,
+    }).isRequired,
```
zburke, line 410: Thanks for providing this more precise description of the shape of location. `PropTypes.object` borders on useless 🙄 . If you're going to do this, however, the new definition has to be complete: `pathname` is missing and now causes a lint error on line 1033.

zburke (changes requested):
Looks good but we need to clean up the newly introduced lint error.

src/routes/RequestsRoute.js
artem-blazhko: Thank you @zburke. Fixed

artem-blazhko:
Hello @zburke. Could you check it one more time?

zburke (approved):
LGTM; thank you! 🤝

PR #1004, opened 2023-03-08
UIREQ-906: Use camel case notation for all data-testid

Dmitriy-Litvinenko:
## Purpose
- Use camel case notation for all data-testid
- Remove magic strings for testIds and labelIds

## Refs
https://issues.folio.org/browse/UIREQ-906

### DIFF
--- a/src/ItemsDialog.test.js
+++ b/src/ItemsDialog.test.js
@@ -32,13 +32,17 @@ jest.mock('./components', () => ({
   )),
 }));
 
+const testIds = {
+  loading: 'loading',
+};
+const labelIds = {
+  selectItem: 'ui-requests.items.selectItem',
+  instanceItems: 'ui-requests.items.instanceItems',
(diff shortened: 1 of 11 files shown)


src/ItemsDialog.test.js
```diff
@@ -33,6 +33,9 @@ jest.mock('./components', () => ({
 }));
 
 describe('ItemsDialog', () => {
+  const testIds = {
+    loading: 'loading',
+  };
```
artem-blazhko, line 38: Could you move it above `describe` to be consistent with test writing approaches?

artem-blazhko (approved):

src/ItemsDialog.test.js
Dmitriy-Litvinenko: Fixed

s3fs (approved):

PR #1028, opened 2023-04-26
UIREQ-836: Split RequestForm component into three smaller parts.

artem-blazhko:
## Purpose
Split RequestForm component into smaller components each of which is responsible for some specific part of the form:
- item information
- instance information
- request information
- requester information

And as a result, it will be easier to support and extend functionality related to the Request form.
 
## Refs
[UIREQ-836](https://issues.folio.org/browse/UIREQ-836)

## Notes
- In general, the functionality was not changed
- All tests should be implemented in the scope of separated tickets (UIREQ-949, UIREQ-950, UIREQ-951, UIREQ-952)


Dmitriy-Litvinenko (approved):

src/components/InstanceInformation/InstanceInformation.js
Terala-Priyanka, line 1: Could you please convert the class based component into function and hooks based as applicable?

src/components/RequesterInformation/RequesterInformation.js
Terala-Priyanka, line 1: Could you please convert the class based component into function and hooks based as applicable?

Terala-Priyanka (changes requested):

src/RequestForm.js
Terala-Priyanka, line 1: Could you please convert the class based component into function and hooks based as applicable?

src/components/RequesterInformation/RequesterInformation.js
artem-blazhko: Answered above

src/RequestForm.js
artem-blazhko: Answered above

Dmitriy-Litvinenko:
> Hi @Terala-Priyanka and @artem-blazhko As far as I know we do not have a strict rule to use functional components instead of classes across the project but this is kinda best practice. Based on the description of the task, I see that it was necessary to break the component into smaller components to make it easier to support and scale, not to refactor the RequestForm to be a functional component (it can be risky and should be well tested to avoid regressions) and it's not a new functionality. I see that author of the task is @Dmitriy-Litvinenko so it's fair to ask him if this PR meets the expectations. And if you started to refactor and improve your code, I would suggest creating a tech-dept to convert it to functional components if it's really needed.

Hello @mariia-aloshyna  thank you. As you said, we don't have rules for using only functional components. The essence of this task is to decompose this component, which will further contribute to the improvement/refactoring code and improvement of code coverage. Refactoring everything at once increases the likelihood of breaking part of the business logic. Also, in the case of a full refactoring, we will need a complete regression testing of the entire module and etc. We have already created tasks for this components to cover or improve coverage of Jest/RTL test and refactor if necessary, which we think is safer. Task UIREQ-836 completed in full.
CC: @artem-blazhko , @Terala-Priyanka

artem-blazhko:
Hello @Terala-Priyanka 
Could you please take a look one more time at this PR?

BogdanDenis (approved):

Terala-Priyanka (approved):

PR #1030, opened 2023-04-26
UIREQ-879: UI tests replacement with RTL/Jest for src/PositionLink.js

Dmitriy-Litvinenko:
## Purpose
UI tests replacement with RTL/Jest for src/PositionLink.js

## Refs
https://issues.folio.org/browse/UIREQ-879

### DIFF
--- a/src/PositionLink.test.js
+++ b/src/PositionLink.test.js
@@ -1,0 +1,90 @@
+import React from 'react';
+import {
+  render,
+  screen,
+} from '@testing-library/react';
+
+import {
+  requestStatuses,
+} from './constants';
+
+import '../test/jest/__mock__';
+
+import PositionLink from './PositionLink';
+
+jest.mock('react-router-dom', () => ({
+  Link: jest.fn(({ to, children }) => <a href={to}>{children}</a>),
+}));
+
+const testIds = {
+  positionLink: 'positionLink',
+};
+const labelIds = {
+  requests: 'ui-requests.requests',
+  viewRequestsInQueue: 'ui-requests.actions.viewRequestsInQueue',
+};
+
+describe('PositionLink', () => {
+  const baseConfig = {
+    request: {},
+    isTlrEnabled: true,
+  };
+  const normalConfig = {
+    request: {
+      status: requestStatuses.AWAITING_DELIVERY,
+      position: 1,
+      instanceId: 'instanceId',
+      itemId: 'itemId',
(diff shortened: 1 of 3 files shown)


src/PositionLink.test.js
```diff
@@ -0,0 +1,90 @@
...
+
+      it('should render link position', () => {
+        expect(screen.getByText(`${normalConfig.request.position} (${labelIds.requests})`)).toBeInTheDocument();
+      });
+
+      it('should render view requests in queue', () => {
+        expect(screen.getByText(labelIds.viewRequestsInQueue)).toBeInTheDocument();
+      });
+
+      it('should render view requests in queue', () => {
```
artem-blazhko, line 71: Could you use unique test names?
In the current situation, only one test will be executed twice and another one will be ignored.

Dmitriy-Litvinenko: Fixed

artem-blazhko (approved):

src/PositionLink.test.js
```diff
@@ -0,0 +1,90 @@
...
+        expect(screen.getByText(`${normalConfig.request.position} (${labelIds.requests})`)).toBeInTheDocument();
+      });
+
+      it('should render view requests in queue', () => {
+        expect(screen.getByText(labelIds.viewRequestsInQueue)).toBeInTheDocument();
+      });
+
+      it('should create correct href', () => {
+        const filters = 'requestStatus.Open - Not yet filled,requestStatus.Open - Awaiting pickup,requestStatus.Open - Awaiting delivery,requestStatus.Open - In transit';
+        const href = `/requests?filters=${filters}&query=${isTlrEnabled ? 'instanceId' : 'itemId'}&sort=Request Date`;
```
BogdanDenis, line 73: IMO tests shouldn't have any logic at all. As soon as tests contain some kind of logic it's much more difficult to believe that tests themselves don't contain any bugs
I know that it means some duplication, but then you have more confidence in tests
If you think that it's ok to have logic in this case - please rename `positionLink` to something like `runSuiteWithParam` so it's easier to understand what's happening

BogdanDenis (changes requested):

src/PositionLink.test.js
Dmitriy-Litvinenko: Hello @BogdanDenis Thank you. Could you please re-review?

PR #1036, opened 2023-05-09
UIREQ-954: Change spelling of fulfillmentPreference

artem-blazhko:
## Purpose
- Update `circulation` okapi interface to `14.0` version.
- use `fulfillmentPreference` param from the server instead of `fulfilmentPreference`
- find and fix all typos related to word `fulfillment`

## Refs
[UIREQ-954](https://issues.folio.org/browse/UIREQ-954)


package.json
```diff
@@ -29,7 +29,7 @@
     "queryResource": "query",
     "okapiInterfaces": {
       "cancellation-reason-storage": "1.1",
-      "circulation": "13.0",
+      "circulation": "14.0",
```
zburke, line 32: Dropping support for an interface is a breaking change. You must also change the module version on line 3 to `9.0.0`.

zburke (changes requested):
The actual changes look good, but the module version needs to get bumped before this can be merged.

zburke (approved):

PR #1060, opened 2023-05-26
UIREQ-946 JEST/RTL test cases for RequestQueueRoute

KetineniM:
UIREQ-946 JEST/RTL test cases for RequestQueueRoute
URL: https://issues.folio.org/browse/UIREQ-946


cchitneedi (approved):

src/routes/RequestQueueRoute.test.js
```diff
@@ -0,0 +1,228 @@
...
+        history={mockHistory}
+        match={mockMatch}
+        location={mockLocation}
+      />
+    );
+    expect(mockMutator.requests.reset).toHaveBeenCalledTimes(1);
+    expect(mockMutator.requests.GET).toHaveBeenCalledTimes(1);
+    expect(mockMutator.requests.GET).toHaveBeenCalledWith({
+      path: 'circulation/requests/queue/item/2',
+    });
```
artem-blazhko, line 175: Could you create separate tests for each "expect"?

artem-blazhko (changes requested):

Dmitriy-Litvinenko (commented):
Could you please fix feedbacks?

src/routes/RequestQueueRoute.test.js
```diff
@@ -0,0 +1,219 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import RequestQueueRoute from './RequestQueueRoute';
+
+jest.mock('@folio/stripes/core', () => ({
+  stripesConnect: jest.fn(Component => props => <Component {...props} />)
+}));
```
artem-blazhko, line 8: We already have it in `test/jest/__mock__/stripesCore.mock.js` file

src/routes/RequestQueueRoute.test.js
```diff
@@ -0,0 +1,219 @@
...
+      <RequestQueueRoute
+        resources={mockResources}
+        mutator={mockMutator}
+        history={mockHistory}
+        location={mockLocation}
+        match={mockMatch}
+      />
+    );
+
+    expect(mockMutator.requests.GET).toHaveBeenCalled();
```
artem-blazhko, line 138: Seems that it is redundant and it is a part of the next test.

src/routes/RequestQueueRoute.test.js
```diff
@@ -0,0 +1,219 @@
...
+        resources={mockResources}
+        mutator={mockMutator}
+        history={mockHistory}
+        location={mockLocation}
+        match={mockMatch}
+      />
+    );
+    expect(container).toBeInTheDocument();
+  });
+  it('should call setTlrSettings on mount if configs have loaded', () => {
```
artem-blazhko, line 112: Name of the test does not fit test body.
In this test you check that `requests.GET` is called after rerender and is not called after mount.

artem-blazhko (changes requested):

Dmitriy-Litvinenko:
Could you please fix feedbacks?

src/routes/RequestQueueRoute.test.js
artem-blazhko: Name of the test does not fit test body.
In this test you check that requests.GET is called after rerender but not after the first render (mount).

Dmitriy-Litvinenko (approved):

artem-blazhko (approved):

PR #1071, opened 2023-07-04
UIREQ-972: Remove redundant ariaLabel prop

artem-blazhko:
## Purpose
When a user opens the request form to create or edit a request we can see a console error related to the missing param for `formatMessage`.
In our case, we had an unnecessary `ariaLabel` attribute which contains the result of execution of `formatMessage`.
Currently `MultiColumnList` component does not support `ariaLabel` property.

## Approach
I have removed `ariaLabel` attribute which is unnecessary for `MultiColumnList` component and as a result I removed `formatMessage({ id: 'ui-requests.items.instanceItems' })` that does not contain the required `title` param.

## Refs
[UIREQ-972](https://issues.folio.org/browse/UIREQ-972)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 * Cover ItemInformation by jest/RTL tests. Refs UIREQ-949.
 * Cover InstanceInformation by jest/RTL tests. Refs UIREQ-950.
 * Fix inconsistency in RTL/Jest tests. Refs UIREQ-979.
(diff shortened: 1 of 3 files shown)


Dmitriy-Litvinenko (approved):

zburke (approved):
LGTM

I appreciate the clear explanation of why removal of the (unused) prop is the correct solution. Thank you!

PR #1086, opened 2023-08-21
UIREQ-977: Ensure Only Available Service Points Display When Placing a New Request

artem-blazhko:
## Purpose
Add functionality to get service points that depend on the request type.
If the request type is selected - a dropdown with service points becomes visible and contains service points that allowed for the selected request type.

## Refs 
[UIREQ-977](https://issues.folio.org/browse/UIREQ-977)

### DIFF
--- a/src/components/FulfilmentPreference/FulfilmentPreference.js
+++ b/src/components/FulfilmentPreference/FulfilmentPreference.js
@@ -53,7 +53,7 @@ const FulfilmentPreference = ({
         {deliveryLocations.map(({ value, label }) => <option key={value} value={value}>{label}</option>)}
       </Field>
     );
-  }
+  };
   const renderPickupServicePoint = () => {
     const selectedRequestType = isEditForm ? request.requestType : values.requestType;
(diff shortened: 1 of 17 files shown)


src/components/FulfilmentPreference/FulfilmentPreference.js
```diff
@@ -0,0 +1,177 @@
+import PropTypes from 'prop-types';
+import { Field } from 'react-final-form';
+import { FormattedMessage } from 'react-intl';
+
+import {
+  Col,
+  Row,
+  Select,
+} from '@folio/stripes-components';
```
BogdanDenis, line 9: stripes-components, stripes-smart-components, stripes-core imports should be imported like `@folio/stripes/components` etc

BogdanDenis (changes requested):

src/components/FulfilmentPreference/FulfilmentPreference.js
artem-blazhko: Fixed

BogdanDenis (approved):

PR #1090, opened 2023-08-31
UIREQ-918: Correctly escape CQL special characters in tag API queries

Dmitriy-Litvinenko:
## Purpose
Correctly escape CQL special characters in tag API queries

## Refs
https://issues.folio.org/browse/UIREQ-918


zburke (approved):
A better CHANGELOG statement would describe what the code now does, e.g. "Correctly escape CQL special characters in tag API queries", rather than stating what the bug was. Thanks for the thorough tests!

PR #1093, opened 2023-09-07
UIREQ-564: Hit Enter or Search should move focus to the Results List pane header

Dmitriy-Litvinenko:
## Purpose
Hit Enter or Search should move focus to the Results List pane header 

## Refs
https://issues.folio.org/browse/UIREQ-564

### DIFF
--- a/src/routes/RequestsRoute.test.js
+++ b/src/routes/RequestsRoute.test.js
@@ -1,3 +1,5 @@
+import React from 'react';
+
 import {
   render,
   screen,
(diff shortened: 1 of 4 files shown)


UladzislauKutarkin (approved):

src/routes/RequestsRoute.test.js
```diff
@@ -1,3 +1,5 @@
+import React from 'react';
```
artem-blazhko, line 1: It is not necessary

artem-blazhko (approved):

PR #1096, opened 2023-09-08
UIREQ-1020: Prefer @folio/stripes exports to private paths when importing components

Dmitriy-Litvinenko:
## Purpose
Prefer @folio/stripes exports to private paths when importing components

## Refs
https://issues.folio.org/browse/UIREQ-1020


artem-blazhko (approved):

zburke (approved):
❤️

PR #1164, opened 2024-05-20
UIREQ-1091: Use appropriate endpoint to get item/instance information

artem-blazhko:
## Purpose
Use newly created `circulation/items-by-instance` endpoint to get information about instance and item.
The main purpose of using that endpoint is presence of `tenantId` for each item and instance that will be used a bit later for other server requests.

## Refs
https://folio-org.atlassian.net/browse/UIREQ-1091


UladzislauKutarkin (approved):

Dmitriy-Litvinenko (approved):

zburke (commented):
The PR description mentions a "newly created endpoint" which implies there needs to be a version bump to the `circulation` okapiInterface, but it is unchanged here. If this endpoint _is_ only available on a new version, this is a breaking change as-implemented. The alternative would be to restructure your manifests and your code with `<IfInterface>` etc. to handle the different versions separately. I would expect that to get pretty messy, so the breaking change is probably the simpler option. Since this is an app (i.e. the top of the dependency hierarchy, with nothing else depending on it except the platform) there are no real tradeoffs to a breaking change.

zburke:
> this PR created to UXPROD-4559-ECS feature branch 

Fine, but irrelevant. If this work uses a new interface or a new version of an existing interface, those changes must be part of this PR unless they are already part of the destination branch.

artem-blazhko:
> > this PR created to UXPROD-4559-ECS feature branch
> 
> Fine, but irrelevant. If this work uses a new interface or a new version of an existing interface, those changes must be part of this PR unless they are already part of the destination branch.

Hello @zburke 
I have added new interface. 
Do you mind if I merge this PR to feature branch?

zburke (approved):
Thank you for updating `version` to a new major version. Consider calling out this change in the CHANGELOG to make it easy to identify why the version bumped to `10` instead of `9.2` with an entry like 
```
*BREAKING* Use `circulation/items-by-instance` ...
```
and updating the version in the CHANGELOG as well.

artem-blazhko:
> Thank you for updating `version` to a new major version. Consider calling out this change in the CHANGELOG to make it easy to identify why the version bumped to `10` instead of `9.2` with an entry like
> 
> ```
> *BREAKING* Use `circulation/items-by-instance` ...
> ```
> 
> and updating the version in the CHANGELOG as well.

Done

PR #1184, opened 2024-08-19
UIREQ-1127, UIREQ-1128: Hide Duplicate and Move action buttons in ECS env with mod-tlr enabled

artem-blazhko:
## Purpose
For ECS environments with mod-tlr enabled "Duplicate" and "Move request" buttons from action dropdown should be hidden.

## Approach
"Duplicate" and "Move request" should be hidden if request is primary, if request is secondary, if ECS TLR setting is enabled.
New functionality was implemented based on those three conditions. 

## Refs
[UIREQ-1127](https://folio-org.atlassian.net/browse/UIREQ-1127)
[UIREQ-1128](https://folio-org.atlassian.net/browse/UIREQ-1128)


Dmitriy-Litvinenko (approved):

zburke (approved):
Code looks great. It would be helpful, especially in RequestsRoute, to have some comments to explain what a "multi-data tenant" is, what the impact of being multi-tenant with TLR enabled is, etc.

PR #1219, opened 2024-10-30
UIREQ-1172: Update ui-plugin-find-instances to version 8.0.0

Dmitriy-Litvinenko:
## Purpose
Update ui-plugin-find-instances to version 8.0.0

# Refs
https://issues.folio.org/browse/UIREQ-1172

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 * Navigate to first page when saving the pick slip print log from beyond the first page. Refs UIREQ-1145.
 * Also support okapiInterfaces `inventory` `14.0`. Refs UIREQ-1160.
 * Review and cleanup Module Descriptor. Refs UIREQ-1156.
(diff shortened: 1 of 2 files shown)


Terala-Priyanka (approved):
In general, it is a breaking change.
But here it is optional interface dependency. If it is still considered as breaking, please mark it as breaking and bump up the major version of your module

Terala-Priyanka (changes requested):
In general, it is a breaking change.
But here it is optional interface dependency. If it is still considered as breaking, please mark it as breaking and bump up the major version of your module

artem-blazhko (approved):

Dmitriy-Litvinenko:
@Terala-Priyanka Hello on my opinion we can do it in scope of release task (_Increment the version number in package.json, if this has not already been done -- **bumping the major version if there are backwards-incompatible changes**, and the minor version if all changes are backwards-compatible._ from https://github.com/folio-org/stripes/blob/master/doc/release-procedure.md#release-procedure)

Terala-Priyanka (approved):

PR #1234, opened 2024-11-27
UIREQ-1201: Update permission after mod-patron-blocks permission changes

Dmitriy-Litvinenko:
## Purpose
Update permission after mod-patron-blocks permission changes

## Refs
https://issues.folio.org/browse/UIREQ-1201

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * *BREAKING* Migrate to new `mod-circulation-bff` endpoints. Refs UIREQ-1134.
 * Implement feature toggle for ECS and not ECS envs. Refs UIREQ-1171.
 * Hide proxy functionality for ECS with mod-tlr enabled. Refs UIREQ-1185.
+* Update permission after mod-patron-blocks permission changes. Refs UIREQ-1201.
 
 ## [10.0.2] (https://github.com/folio-org/ui-requests/tree/v10.0.2) (2024-11-22)
 [Full Changelog](https://github.com/folio-org/ui-requests/compare/v10.0.1...v10.0.2)
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
         "displayName": "Requests: View, create",
         "subPermissions": [
           "ui-requests.view",
-          "automated-patron-blocks.collection.get",
+          "patron-blocks.automated-patron-blocks.collection.get",
           "circulation.requests.item.post",
           "circulation.requests.allowed-service-points.get",
           "circulation-bff.requests.allowed-service-points.get",
--- a/package.json
+++ b/package.json
@@ -124,6 +124,7 @@
         "displayName": "Requests: View, create",
         "subPermissions": [
           "ui-requests.view",
+          "automated-patron-blocks.collection.get",
           "patron-blocks.automated-patron-blocks.collection.get",
           "circulation.requests.item.post",
           "circulation.requests.allowed-service-points.get",


Dmytro-Melnyshyn (approved):

zburke (approved):
Was there an accompanying change to the `automated-patron-blocks` interface version (or some other interface)? If there wasn't, there should have been 😬 and since you're dropping support for the old permission by removing `automated-patrong-blocks.collection.get`, that should be mentioned as a breaking change in the CHANGELOG. Alternatively, leave both permissions in place and update the interface dependency.

In NPM-land, leading zeros in a version string cause the minor version to behave like a major from the point of view of semver constraints, e.g. `^6.1.0` will give you `>= 6.1.0, < 7.0.0`, but `^0.6.1` will give you `>= 0.6.1, < 0.7.0`. I don't know how we handle this with respect to okapi interface versions, but it feels to me like this should have been accompanied by a change like
```
      "automated-patron-blocks": "0.1 0.2",
```

Oooof. There's a lot that's going on with this PR that is outside your control. Given that, and all you can do is react to the eco system you're living in, I will approve this. But I think you should consider leaving both permissions in place in order to maintain backwards compatibilty, even though the interface version didn't change the way it should have.

Dmitriy-Litvinenko:
> Was there an accompanying change to the `automated-patron-blocks` interface version (or some other interface)? If there wasn't, there should have been 😬 and since you're dropping support for the old permission by removing `automated-patrong-blocks.collection.get`, that should be mentioned as a breaking change in the CHANGELOG. Alternatively, leave both permissions in place and update the interface dependency.
> 
> In NPM-land, leading zeros in a version string cause the minor version to behave like a major from the point of view of semver constraints, e.g. `^6.1.0` will give you `>= 6.1.0, < 7.0.0`, but `^0.6.1` will give you `>= 0.6.1, < 0.7.0`. I don't know how we handle this with respect to okapi interface versions, but it feels to me like this should have been accompanied by a change like
> 
> ```
>       "automated-patron-blocks": "0.1 0.2",
> ```
> 
> Oooof. There's a lot that's going on with this PR that is outside your control. Given that, and all you can do is react to the eco system you're living in, I will approve this. But I think you should consider leaving both permissions in place in order to maintain backwards compatibilty, even though the interface version didn't change the way it should have.

@zburke Hello "Was there an accompanying change to the automated-patron-blocks interface version (or some other interface)?" no


> Was there an accompanying change to the `automated-patron-blocks` interface version (or some other interface)? If there wasn't, there should have been 😬 and since you're dropping support for the old permission by removing `automated-patrong-blocks.collection.get`, that should be mentioned as a breaking change in the CHANGELOG. Alternatively, leave both permissions in place and update the interface dependency.
> 
> In NPM-land, leading zeros in a version string cause the minor version to behave like a major from the point of view of semver constraints, e.g. `^6.1.0` will give you `>= 6.1.0, < 7.0.0`, but `^0.6.1` will give you `>= 0.6.1, < 0.7.0`. I don't know how we handle this with respect to okapi interface versions, but it feels to me like this should have been accompanied by a change like
> 
> ```
>       "automated-patron-blocks": "0.1 0.2",
> ```
> 
> Oooof. There's a lot that's going on with this PR that is outside your control. Given that, and all you can do is react to the eco system you're living in, I will approve this. But I think you should consider leaving both permissions in place in order to maintain backwards compatibilty, even though the interface version didn't change the way it should have.

Hello @zburke 
1. "Was there an accompanying change to the `automated-patron-blocks` interface version (or some other interface)?" no 
2. "automated-patron-blocks.collection.get" added

PR #1273, opened 2025-02-21
UIREQ-1261: Ramsons - Requests - #Copies not incrementing with all print actions

s3fs:
Filter data sent to track amount of times a request has been printed for duplicates.

https://folio-org.atlassian.net/browse/UIREQ-1261


zburke (approved):
Code change LGTM. 

It would be nice to see an updated test for this. The function is pretty simple so it should not take much effort.

mariia-aloshyna (approved):

PR #1286, opened 2025-03-06
UIREQ-1261: Don't include items without barcode in PDF for pick slips

s3fs:
https://folio-org.atlassian.net/browse/UIREQ-1261,


Dmitriy-Litvinenko (changes requested):
function convertToSlipData should convert to slip data and this all. If you need exclud something please do it in seporate function before transfer data to convertToSlipData. Also this is re-usebel functiion that will be moved from curent modules.

zburke (commented):
IIUC @Dmitriy-Litvinenko's comments, he would rather see `.filter()` applied before passing `sources` into this function in order to keep its implementation simple. That's sensible advice, though there is a solid argument to be made that this function should be liberal/tolerant about the kind of data it is able to receive. 

I don't have a stake in the outcome here, but will note that if you decide `convertToSlipData` _should_ handle the filter, then 
1. I also agree with @Dmitriy-Litvinenko that we should make this change in the shared function in stripes-utils instead
2. It should be covered by a test

zburke (approved):

PR #1290, opened 2025-03-13
UIREQ-1275: User gets an error when trying to create a Title level request and select an Item without barcode

artem-blazhko:
## Purpose
Fix issue when creating a request with an item that does not have a barcode.

## Refs
[UIREQ-1275](https://folio-org.atlassian.net/browse/UIREQ-1275)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * Add optionalOkapiInterfaces `circulation-bff-loans` `1.0`. Refs UIREQ-1273.
 * For consistency searching use == instead of = for requester barcode. Refs UIREQ-640.
 * Add missed sub-permissions for "Requests: View" permission. Refs UIREQ-1217.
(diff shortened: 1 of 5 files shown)


zburke (approved):
Code changes LGTM but documentation is poor.

Fix your PR title and CHANGELOG entry. They describe the bug instead of the solution.

Explain in your PR description why setting `isItemIdRequest` resolves the problem. The documentation here describes the problem but not the solution, forcing a reviewer (or some future-person spelunking through git history) to re-triage this problem all over again, trying to understand what the bug is and then how this change resolves it. We can learn about the UI problem from the Jira ticket. We can learn about what changed from `git diff`. We need you to explain what the code problem was, and _why_ this change fixed it. 

It would be polite to clean up the lint issues in the files you are changing, even if they are not your errors. 

I very strongly oppose the use of the `deprecated` folder. It's been present through two major-version releases. Eliminate it as soon a possible.

Dmitriy-Litvinenko (approved):

PR #1296, opened 2025-04-08
UIREQ-1294: [Retrieval SP filter] Move related code to make the filter enabled for ecs/non-ecs UI

s3fs:
Move related code to make the Retrieval SP filter enabled for ecs/non-ecs UI.

https://folio-org.atlassian.net/browse/UIREQ-1294

### DIFF
--- a/src/deprecated/routes/RequestsRoute/RequestsRoute.js
+++ b/src/deprecated/routes/RequestsRoute/RequestsRoute.js
@@ -942,7 +942,7 @@ class RequestsRoute extends React.Component {
         const { addressLine1, city, region, postalCode, countryId } = record.deliveryAddress;
         record.deliveryAddress = `${addressLine1 || ''} ${city || ''} ${region || ''} ${countryId || ''} ${postalCode || ''}`;
       }
-      if (record.item.retrievalServicePointName) {
+      if (record?.item?.retrievalServicePointName) {
         record.retrievalServicePointName = record.item.retrievalServicePointName;
       }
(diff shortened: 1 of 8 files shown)


src/deprecated/routes/RequestsRoute/RequestsRoute.js
```diff
@@ -939,6 +942,9 @@ class RequestsRoute extends React.Component {
         const { addressLine1, city, region, postalCode, countryId } = record.deliveryAddress;
         record.deliveryAddress = `${addressLine1 || ''} ${city || ''} ${region || ''} ${countryId || ''} ${postalCode || ''}`;
       }
+      if (record.item.retrievalServicePointName) {
```
Dmitriy-Litvinenko, line 945: Hello. Could you please correct me if i am wrong, but we will have the same issue (https://folio-org.atlassian.net/browse/UIREQ-1292) here?
Could you please clarify maybe we will need change code (just for example `record?.item?.retrievalServicePointName`) to or apply this changes after fix bug?

Dmitriy-Litvinenko (changes requested):

src/deprecated/routes/RequestsRoute/RequestsRoute.js
s3fs: Good catch. Fixed.

Dmitriy-Litvinenko (approved):

UladzislauKutarkin (approved):

PR #1314, opened 2025-06-11
UIREQ-1312: Display Unknown user instead of link with undefined when a user record is deleted

Dmitriy-Litvinenko:
## Purpose
When a patron record is deleted there is still a link in the requests details page

## Refs
https://folio-org.atlassian.net/browse/UIREQ-1312

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * Replace moment with day.js. Refs UIREQ-1291.
 * Reduce count of eslint errors after update eslint-config-stripes. Refs UIREQ-1289.
 * Display loan type and "For use at location" loan-policy setting in new-request form. Fixes UIREQ-1282.
-* Display Unknown user instead of undefined when a user record is deleted. Refs UIREQ-1312.
+* Display Unknown user instead of link with undefined when a user record is deleted. Refs UIREQ-1312.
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * Replace moment with day.js. Refs UIREQ-1291.
 * Reduce count of eslint errors after update eslint-config-stripes. Refs UIREQ-1289.
 * Display loan type and "For use at location" loan-policy setting in new-request form. Fixes UIREQ-1282.
+* When a patron record is deleted there is still a link in the requests details page. Refs UIREQ-1312.
```
mariia-aloshyna, line 9: I guess it is better to describe the change the way "what has been done", instead of just describing a problem. Like "Display Unknown user instead of a link when a patron record is deleted"

src/utils.js
```diff
@@ -66,7 +66,7 @@ export function userHighlightBox(title, name, id, barcode) {
             {title}
           </Headline>
           <div>
-            {recordLink}
+            {recordLink || <FormattedMessage id="ui-requests.errors.user.unknown" />}
```
Dmytro-Melnyshyn, line 69: The results pane also shows `undefined` in the `Printed` column.

CHANGELOG.md
Dmitriy-Litvinenko: Fixed

Dmytro-Melnyshyn (approved):

PR #1316, opened 2025-07-07
UIREQ-1296: Add missing subPermissions for Requests: View, create permission

Dmitriy-Litvinenko:


BogdanDenis (approved):

zburke (approved):
LGTM. 

Consider a CHANGELOG entry to make it easy to see when this bug was resolved if this PR patches a bug that snuck into production. If "make API requests to mod-search" will be a new feature in the upcoming release, then it's not really necessary since it's just a follow up to that PR. Either way, consider linking to the PR where this bug was introduced so we can help the folks who were involved in that change avoid this kind of error in the future.

PR #1321, opened 2025-08-05
[UIREQ-1325] Show loan type of requested item

MikeTaylor:
The same component `<ItemDetail>` is used to show information about the requested item both while creating a new request and when viewing an existing one.

In the former case, we have full information about the item, because we fetch it as part of the process of making a new request. But in the latter case, we have only the stub item that comes as part of the request object. That's not enough information because we need to display fields such as `loanType` that are not included in the stub item.

This PR modifies the flow of the second case so it goes via a facade component `<FullItemDetail>` which fetches the full item and passes its information through. (Until the item has been fetched, it just passes the stub item though, so that most but not all of the item-detail fields can be rendered immediately, and those that are only in the full item quietly appear on the re-rended when the fetch is done.)

### DIFF
--- a/src/deprecated/components/ViewRequest/FullItemDetail.test.js
+++ b/src/deprecated/components/ViewRequest/FullItemDetail.test.js
@@ -1,8 +1,8 @@
 // FullItemDetail.test.js
 import React from 'react';
-import { render, screen, waitFor } from '@testing-library/react';
-import FullItemDetail from './FullItemDetail';
+import { render, screen, waitFor } from '@folio/jest-config-stripes/testing-library/react';
 import { useOkapiKy } from '@folio/stripes/core';
+import FullItemDetail from './FullItemDetail';
 import ItemDetail from '../../../ItemDetail';
 
 // Mock the useOkapiKy hook


alexanderkurash (commented):
Hi Mike. We discussed UIREQ-1325 with Dmitriy and our POs and I think we all agree that this should be done in mod-circulation rather than ui-requests.
Let me quickly describe how we manage request details info. 
UI makes a call to `GET /circulation/requests` which is not a storage endpoint. It's an endpoint in mod-circulation that gets info from a lot of different places, like inventory and circulation storage, and then combines it into a single response (basically, a BFF concept). This way, UI gets all information it needs to render request details with one call to the BE, or at least it minimizes the number of calls. When we need to add more info, we just add it on the mod-circulation side by changing `RequestRepresentation.extendedRepresentation()` method. In this case, all is needed is to add loan type to the `RequestRepresentation.addItemProperties()` method. Item info is already there, so probably all is needed is to add something like
```
write(itemSummary, "loanTypeName", item.getLoanTypeName());
```
(or whatever the field name we want to see on the FE side) to  `addItemProperties()` method. Of course, unit tests need to be extended to include this field, but I think for a BE developer it's 1-2 days of work max.

MikeTaylor:
Thanks, @alexanderkurash. The issue, though, is not how much effort it will take to do this work, but what the elapsed time will be. We need this now for a customer. A day or two's delay is tolerable, but if it stretches to weeks that will be bad.

PR #1323, opened 2025-08-07
[UIREQ-1328] Display loan type on the Request Detail page

MikeTaylor:
(As well as New Request page, which worked before and still does.)

All we're doing here is looking at the field that's been added to the request record's `item` subrecord in https://github.com/folio-org/mod-circulation/pull/1607 so that when we're viewing an existing request, we're able to show the loan-type. We also still look at the fields we checked before (`item.temporaryLoanType?.name` and `item.permanentLoanType?.name`), as these are still where the information shows up in the item record itself during the process of creating a new request.


MikeTaylor:
@zburke @NikitaSedyx @mariia-aloshyna @vashjs @usavkov-epam Could someone please do a second review on this very simple PR? I want to get it merged so I can build on it without having to manage multiple branches.

MikeTaylor:
@zburke @NikitaSedyx @mariia-aloshyna @vashjs @usavkov-epam Could someone please do a second review on this very simple PR? I want to get it merged so I can build on it without having to manage multiple branches.

PR #1369, opened 2026-05-25
UIREQ-1363: Fix Year field to correctly parse and display multiple years

Dmitriy-Litvinenko:
## Purpose
Fix Year field to correctly parse and display multiple years

# Refs
https://folio-org.atlassian.net/browse/UIREQ-1363

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for ui-requests
 
+## [13.1.0] IN_PROGRESS
+
+* Fix "Year" field to correctly parse and display multiple years. Refs UIREQ-1363.
+
 ## [13.0.1] (https://github.com/folio-org/ui-requests/tree/v13.0.1) (2026-05-06)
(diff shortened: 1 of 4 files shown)


BogdanDenis (approved):

OleksandrHladchenko1 (approved):

zburke (approved):
This is the right fix for the wrong problem. It is the right fix because it does what the ticket asks (fine, approved) but this change doubles down on an undocumented convention for supplying structured data in a free-text field. Is the PO aware of this glaring inadequacy? There is no front-end validation for this structured data. Did you file a follow-up bug? There is no back-end validation for this structured data. Did you file a follow-up bug?

Secondly, "fix year field" does not describe the problem or how this change solves it. Don't make me read your code to figure out what is happening and why. 

This would have been a great opportunity to add some documentation to `getFormattedYears()`. What is the shape of the arguments going in and the string coming out?

